PR #2, opened 2018-11-08
Prepare application for deployment with Mina, Puma and NGinx

arneboockmeyer:


config/deploy.rb
```diff
@@ -0,0 +1,100 @@
...
+#   domain       - The hostname to SSH to.
+#   deploy_to    - Path to deploy into.
+#   repository   - Git repo to clone from. (needed by mina/git)
+#   branch       - Branch name to deploy. (needed by mina/git)
+
+set :application_name, 'VM-Managment Portal'
+set :domain, 'vm-swt-hrmt-master.eaalab.hpi.uni-potsdam.de'
+set :deploy_to, '/var/www/vm-swt-hrmt-master.eaalab.hpi.uni-potsdam.de'
+set :repository, 'git@github.com:hpi-swt2/vm-portal.git'
+set :branch, 'deploy'
```
arneboockmeyer: Change this to `master` and think about multistaging

.gitignore
```diff
@@ -30,5 +30,9 @@
 # Ignore master key for decrypting credentials and more.
 /config/master.key
 
+# Ignore config files
+/config/database.yml
```
chrisma: Why should changes to the database.yml not be checked into version control?

Gemfile
```diff
@@ -8,8 +8,9 @@ ruby '2.5.0'
 # Gems to instalL: faker, shoulda matchers
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '~> 5.2.1'
-# Use sqlite3 as the database for Active Record
-gem 'sqlite3'
+# Use sqlite3 and postgres as the database for Active Record
+gem 'sqlite3' # development and testing
+gem 'pg'      # production
```
chrisma: Should this gem not go into the 'production' group? It is not needed for testing or development environments, right?

Gemfile
```diff
@@ -47,6 +48,12 @@ gem 'jquery-rails'
 # Vsphere access
 gem 'rbvmomi'
 
+# Mina for deployment
+gem 'mina', require: false
+gem 'mina-puma', require: false,  github: 'untitledkingdom/mina-puma'
```
chrisma: Links to the documentation of these projects / a sentence on what they do would be nice.

Gemfile
```diff
@@ -74,6 +81,9 @@ group :development do
   # Much nicer error pages https://github.com/BetterErrors/better_errors
   gem 'better_errors'
   gem 'binding_of_caller'
+  # Mina for deployment
+  #gem 'mina', require: false
+  #gem 'mina-puma', require: false,  github: 'untitledkingdom/mina-puma'
```
chrisma: why are these gems commented out? Can they be removed? Why are they in the 'development' group?

config/deploy.rb
krichly: There is a mina multi-staging gem! We can discuss this after we copied the VM.

.gitignore
krichly: Because it is a defacto standard to ignore the database.yml. Additionally, each deployment server has a specific database configuration, which is located in the shared files folder, which leads to conflicts, if there is another database.yml in the repository.

arneboockmeyer:
Squashed in https://github.com/hpi-swt2/vm-portal/commit/becd9778ddf80ae31f1e5efafddd1366de8c53d6

PR #46, opened 2018-11-15
Handle successful and unsuccessful login attempts

mlichtblau:
Redirect to vm index page after successful login #34 
Display correct error messages after unsuccessful error messages #35 #36

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
+  def after_sign_in_path_for(_resource)
+    vm_index_path
+  end
 end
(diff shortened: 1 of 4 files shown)


mlichtblau:
I think the vm index page does something, which causes the server to respond with an error (only on Travis). We could probably mock the index page to be more independent...

marcfreiheit:
We might ask the vm team about the test?

mlichtblau:
@BjoernDaase have you noticed this problem with Travis? Does your team already have a solution for this?

BjoernDaase:
I personally haven't noticed anything. Maybe someone of you @hpi-swt2/bd-18?

chrisma:
The failing test seems to be `./spec/features/sessions/user_sign_in_spec.rb` which attempts to check whether a success message after successful login is shown.
However, somewhere in the code under test (possibly by redirecting to `/vm`), an attempt to connect to the VMware vSphere center is made, which fails, as Travis CI cannot reach the service as it is not part of the university network.

mlichtblau:
After doing some research I think we can safely delete the test.
The redirect is should probably be tested with a controller test (there you can use the `redirect_to` matcher). In our case the controller is provided by the devise gem and can therefore not be tested.

PR #52, opened 2018-11-19
Add test user for seeding the database

marcfreiheit:
This pull requests enriches the seeds file for a user, which can be used for local development. In order to populate the database with the given user, you have to run 
`rails db:seed`
Then you can login with: `test@test.de` and `123456`.


janwestphal306:
Deprecated because of #117

PR #103, opened 2018-11-26
Add notification page. #63

Adrian-St:
Add overview page for notifications. The notifications page is not yet accessible via the header, for that is another user story in another issue. To look at the new functionality

 - be logged in
 - go to `/notifications/new` and create a notification
 - look at `/notifications` to see all notifications for you

notifications_page


arne-z:
If possible, you should display an empty list or a text saying "You don't have any notifications at the moment" when visiting the page while I have no notifications. You should avoid displaying pages with only a headline and no content as it looks very wierd to the user.

hkohlsaat:
@arne-z please have a look on the explanatory text which is displayed on the notification page, if no notifications are there to be displayed, and let us know, so we can assign reviewers. Thx

hkohlsaat:
# Next steps
- [ ] let arne look over it
- [ ] run rubocop autofix
- [ ] assign reviewers
- [ ] merge

arne-z:
I've looked over it and it is fine so far.

PR #105, opened 2018-11-27
dashboard 

NiklasKoehnecke:
Pull request for #16


arne-z:
I really like the look of this so far, thinking forward, you should name the Placeholder-Box at the top right "Notifications" since that is what will be displayed there later.   
You should also adjust the Dashboard button in the header to link to your new page. I think the second placeholder box on the right will not be needed any time soon, it should likely be removed.  
One thing I would like to note though is that Hosts are NOT the same as Servers. You currently have a list of Hosts at the bottom, this will be replaced by a list of servers, once that one is available at /servers.

PR #107, opened 2018-11-27
18/scaffold/automated vm creation

MaxSchneider1337:


MaxSchneider1337:
@regnujAx Our old approach was not working, because the link_to methods are meant to call javascript and not ruby code. So I tried this way of calling a controller action. Now there is a route for the controller method and we can call it like a normal path. 

We still need to revert some changes (like deleting the helper methods and stop using the helper in the controller etc) and refactor the tests

PR #117, opened 2018-11-29
Login dev gh 38 39

lassesteffen:


skydivin4ng3l:
The current Build failure is caused by the unreachable VM servers by the Travis Servers.
Caused by this test case `spec/features/sessions/user_sign_up_spec.rb`

Locally reproducible when not at HPI and running Rspec, can be solved locally by connecting per VPN.  But not on the Travis server. 

Test should be changed

config/routes.rb
```diff
@@ -6,7 +6,7 @@
 
   get '/server/:id' => 'server#show', constraints: { id: /.*/ }
 
-  devise_for :users, path: 'users'
+  devise_for :users, controllers: { registrations: 'users/registrations' }, path: 'users'
```
skydivin4ng3l: I pulled over the path: 'users' to devise_for

config/routes.rb
```diff
@@ -6,7 +6,7 @@
...
   resources :vm, :server
```
skydivin4ng3l: pulled over from dev the resoures for :vm all cause by a merge conflict

spec/features/sessions/user_sign_up_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  it 'creates a user with its profile data' do
+    page.fill_in 'user[user_profile_attributes][first_name]', with: 'Max'
+    page.fill_in 'user[user_profile_attributes][last_name]', with: 'Mustermann'
+
+    page.fill_in 'user[email]', with: 'max@mustermann.com'
+    page.fill_in 'user[password]', with: '123456'
+    page.fill_in 'user[password_confirmation]', with: '123456'
+
+    page.find('input[type=submit]').click
```
skydivin4ng3l: This submit button will cause the `app/controllers/application_controller.rb` to run
 `def after_sign_in_path_for(_resource)
    vm_index_path
  end`
which then leads to VM index page which tries to collect VM information via the `app/api/vmapi.rb` which will not be successfull out side of HPI network.

mlichtblau: This should be fine now, as it's not intended behaviour for the index page to throw an error when trying to load it.

mlichtblau:
Why do we need a user Profile? Isn't that redundant when we have a user?
PS: Referencing the issue in the PR helps understanding what you're trying to do.

marcfreiheit:
> Why do we need a user Profile? Isn't that redundant when we have a user?
> PS: Referencing the issue in the PR helps understanding what you're trying to do.

Separation of responsibilities. A user should only be used in order to authorise a person. Therefore, it contains a username and password. First, last and full name aren't required for authentication. Therefore, we've decided to map `user` and `user_profile` 1:1. In addition, it might be easier to add more user roles in future.

app/controllers/users/registrations_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+  # def cancel
+  #   super
+  # end
+
+  protected
+
+  # If you have extra params to permit, append them to the sanitizer.
+  def configure_sign_up_params
+    devise_parameter_sanitizer.permit(:sign_up, keys: [
+                                        user_profile_attributes: %i[
```
mlichtblau: What happens here? %i?

app/models/user_profile.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class UserProfile < ActiveRecord::Base
+  # Associations
+  belongs_to :user
+
+  def name
+    "#{first_name} #{last_name}"
```
mlichtblau: Convenient! 👍

app/controllers/users/registrations_controller.rb
lassesteffen: @marcfreiheit

skydivin4ng3l: This is a array of symbols

https://stackoverflow.com/questions/39303737/why-do-you-use-w-in-rails

mlichtblau: Hm ok. Do you use that for a reason? Or just style?

Issue #50, opened 2018-11-16
Issue Templates for Pull Requests and Bug Reports


mlichtblau:
**Beschreibung**

As a developer I want to have Bug Report and Pull Request Templates in order to ensure consistency and simpler issue creation

**Akzeptanzkriterien**

Templates can be used when clicking new issue on Github

**Aufwandsschätzung**

BjoernDaase:
Does anybody already have a template they like? Or specific wishes?

marcfreiheit:
> Does anybody already have a template they like? Or specific wishes?

We will take care of it and create a pull request within the next days. Then you can decide which version you would like to choose.

PR #122, opened 2018-11-30
Add Issue and Pull Request Templates

marcfreiheit:
**Description**

This Pull Request adds templates for both issues and pull requests. It supports consistency and communication throughout all GitHub issues. In addition, these templates guide developers, collaborators and interested GitHub users in creating valuable issues. 

There are templates for ...

**User Stories**

I like to split user stories in three lines in order to easily distinguish between role, desired feature and intention. 
Acceptance Criteria should be added as a checkable list. Then, they can be ticked by our product owners. 
I would suggest to add labels for complexity (currently used `pain points` by some? teams) and priority (also currently used by some teams..). That's easier to filter and more visual than writing it inside the issues itself.

**Bugs**

I would suggest a template containing a small description of the bug, the steps to reproduce the bug, the actual behaviour and the expected behaviour. 
The Pull Requests links to the bug issues and should not duplicate information from that issue.

**Arbitrary Tasks**

Really similar to user stories but less restrictive.

**Common Features**

All templates contain comments which are not rendered when the template is published. Comments are identified by this syntax: `[//]: # (<comment>)`. They describe for each section what the issuer should input.

**Todos**

- [ ] Setup these templates in GitHub. Then, if a new issues is created you can select between all options. 

**Related Issues**

* #50

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,23 @@
+**Summary**
+
+[//]: # (Please provide a details description about the issue you are facing.)
+
+**Steps to Reproduce**
+
+[//]: # (Please provide the steps to reproduce the defect. Try to be as specific as possible in order to avoid ambigious testing.)
+
+1. 
+1. 
+
+**Actual Results**
+
+[//]: # (Describe the results you've got.)
+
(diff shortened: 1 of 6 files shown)


marcfreiheit:
I'm done so far. Ready for review.

PR #123, opened 2018-11-30
Added detail and overview page

### LINKED ISSUE #77: profile-page
**Beschreibung**
As a user, I'd like to see a profile-page, on which I can see all important information about myself. When I click on my name in the header, I want to be redirected there. On the page I find my name, e-mail adress and a possibility to store my ssh key.

**Akzeptanzkriterien**
- [x] Clicking on the name in the header redirects to the profile-page
- [x] E-mail adress is displayed
- [x] Name of user is displayed
- [x] Possible to see if ssh-key was stored or not

**Priorität**
High
**Aufwandsschätzung**

### LINKED ISSUE #84: all users-page
**Beschreibung**
As a user I want a way to see all users that have accounts for the tool. For this I'd like a separate page on which I find a list of all users including their last login-time. I want to have a button in the header by which I can reach the page. By clicking on somebodies' name I want to be redirected to their personal page.

**Akzeptanzkriterien**
- [x] Page for user-overview exists
- [x] Button in header leads to page
- [x] On it is a list of all users and last login-times
- [x] clicking on the username leads to the personal page of the user

**Priorität**

**Aufwandsschätzung**

lassesteffen:
Fixes #84 Fixes #77


chrisma:
@lassesteffen Could this link be helpful? https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#associations

app/models/user_profile.rb
```diff
@@ -2,7 +2,7 @@
 
 class UserProfile < ActiveRecord::Base
   # Associations
-  belongs_to :user
+  belongs_to :user, optional: true
```
skydivin4ng3l: does the optional true here mean that a user does not has to have a userProfile?

app/views/application/_navbar.html.erb
```diff
@@ -15,6 +15,9 @@
       <li class="nav-item">
         <a class="nav-link" href="../servers">Servers</a>
       </li>
+      <li class="nav-item">
+        <%= link_to('Users', users_path, class: "nav-link") %>
```
skydivin4ng3l: #77 does need a button to access the users profile, is this not the path for the index of users?

skydivin4ng3l: I see for #84 its the users and the profile is still missing =)

app/views/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <% @users.each do |user| %>
+      <tr>
+        <td><%= link_to user.email, user_path(user) %></td>
+        <td><%= user.user_profile.first_name %></td>
+        <td><%= user.user_profile.last_name %></td>
+        <td><%= user.current_sign_in_at %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>`
```
skydivin4ng3l: do we need to apply style guides for this page

app/views/users/show.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<p>
+  <strong>Last name:</strong>
+  <%= @user.user_profile.last_name %>
+</p>
+
+
+<p>
+  <strong>SSH-Key:</strong>
+  <%#= @user.user_profile.ssh_key%>
+</p>
```
skydivin4ng3l: do we need to apply style guides for this page

config/routes.rb
```diff
@@ -10,6 +10,7 @@
 
   devise_for :users, controllers: { registrations: 'users/registrations' }, path: 'users'
   resources :vms, :servers
+  resources :users, only: %i[show index]
```
skydivin4ng3l: so this is the creation of users routes but limited to show and index? Cool =)

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+require 'rails_helper'
+
+RSpec.describe UsersController, type: :controller do
+  describe 'GET #index' do
+    it 'returns http success' do
+      get :index
+      expect(response).to have_http_status(:success)
+    end
+  end
+end
```
skydivin4ng3l: should also "show" be tested here?

db/schema.rb
```diff
@@ -30,20 +41,12 @@
     t.datetime 'remember_created_at'
     t.datetime 'created_at', null: false
     t.datetime 'updated_at', null: false
+    t.integer 'sign_in_count', default: 0, null: false
+    t.datetime 'current_sign_in_at'
+    t.datetime 'last_sign_in_at'
+    t.string 'current_sign_in_ip'
+    t.string 'last_sign_in_ip'
     t.index ['email'], name: 'index_users_on_email', unique: true
     t.index ['reset_password_token'], name: 'index_users_on_reset_password_token', unique: true
```
skydivin4ng3l: do we allow password reset when we do HPI login? Should this be removed?

spec/factories/users.rb
```diff
@@ -8,5 +8,6 @@
     email { Faker::Internet.safe_email }
     password { Faker::Internet.password(min_length = 10, max_length = 20, mix_case = true, special_chars = true) }
     password_confirmation { password }
+    association :user_profile, factory: :user_profile
   end
 end
```
skydivin4ng3l: why does this factory looks so different compared to user_profiles?

spec/views/users/index.html.erb_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'users/index.html.erb', type: :view do
+  let(:users) do
+    [
+      FactoryBot.create(:user),
+      FactoryBot.create(:user, email: 'Joe@joe.de')
```
skydivin4ng3l: does these two do the same? except that the second user has a different email?

spec/views/users/show.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'users/show.html.erb', type: :view do
+  let(:user) { FactoryBot.create :user }
+
+  before do
+    assign(:user, user)
```
skydivin4ng3l: does this inject our local :user into the database for the test or why do we need this?

app/helpers/users_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module UsersHelper
```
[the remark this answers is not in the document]
lassesteffen: All strings in this file are frozen and can not be modified. https://stackoverflow.com/questions/37799296/what-does-the-comment-frozen-string-literal-true-do, rubocop complains otherwise

app/models/user_profile.rb
lassesteffen: Yeah it does, not sure if this is super smart. I did it mostly to get my user factory running and work with the profile association. Do you think this harms us somewhere?

lassesteffen: Or actually it does mean that a profile does not automatically needs to belong to a user. So there can be a profile without a User

app/views/application/_navbar.html.erb
lassesteffen: I am not quite sure, what is missing?

app/views/users/show.html.erb
lassesteffen: Good question, I think this one still need styling

app/views/users/index.html.erb
lassesteffen: I will check, thanks!

spec/controllers/users_controller_spec.rb
lassesteffen: I will

db/schema.rb
lassesteffen: Probably, but this is not the scope of this PR. This PR just brings the overview pages, there should be a seperate PR focussed on cleaning stuff up

spec/views/users/index.html.erb_spec.rb
lassesteffen: Exactly

spec/factories/users.rb
lassesteffen: Can you elaborate?

spec/views/users/show.html.erb_spec.rb
lassesteffen: You are right, should actually be enough to just build the user! Good catch

spec/views/users/index.html.erb_spec.rb
chrisma: You might consider using `Faker::Internet.safe_email`
https://github.com/stympy/faker/wiki/Internet

app/models/user_profile.rb
mlichtblau: I'm still not really convinced about the separation of user and user profile. And introducing an optional association seems to only add to the confusion. Is it somehow possible to change that?

app/views/application/_navbar.html.erb
mlichtblau: We should have a button to our own profile page? Maybe display first and last name in the menu bar and when you press it you get redirected to your own profile?

spec/factories/users.rb
skydivin4ng3l: Nevermind github show me just lines 8-12 and i did not understand how it worked without the missing stuff^^

spec/views/users/index.html.erb_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    ]
+  end
+
+  before do
+    assign(:users, users)
```
mlichtblau: Same here?

lassesteffen:
> We should try to keep the pull request as small as possible. For example keeping the detail and overview page in different branches would allow to review code better and merge faster. Just for the next time :)

Sure thing! Thought this was an easy thing and ended up doing something for both, which was not the best idea in hindsigh

app/views/application/_navbar.html.erb
lassesteffen: Done

spec/views/users/show.html.erb_spec.rb
lassesteffen: Actually we still need create, because we access the users_path which is filled with data from the db. I can explain it to you in person later on.

app/views/application/_navbar.html.erb
```diff
@@ -10,10 +10,13 @@
         <a class="nav-link" href="#"><%= fa_icon('desktop', text: "Dashboard") %><span class="sr-only">(current)</span></a>
       </li>
       <li class="nav-item">
-        <a class="nav-link" href="../vms">VMs</a>
+        <%= link_to('VMs', vms_path, class: "nav-link") %>
```
mlichtblau: Nice

Issue #72, opened 2018-11-23
Integration of buttons on the server overview / detail page


schmidtflo:
As a user, I would like to be able to trigger different actions or reach subpages with different buttons on the server overview page / detail page.

- [x] There is a button in the list on every server and on every detail page to access the form for requesting a new timeslot.
- [x] A click on the name of the server leads to the detail page (no button, but simple link)

Priority: medium

PR #124, opened 2018-11-30
Remove Details Button, add Timeslot Button

fewerner:
For issue #72 
- Remove Details Button on Overview Pages and instead link the name to Details Page.

- Add timeslot reservation button for server without destination (does nothing)

### DIFF
--- a/app/views/server/index.html.erb
+++ b/app/views/server/index.html.erb
@@ -36,6 +36,7 @@
     <th scope="col" data-sortable="true">Status</th>
     <th scope="col" data-sortable="true">Running Since</th>
     <th scope="col">VMs</th>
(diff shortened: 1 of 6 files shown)


app/views/server/index.html.erb
```diff
@@ -36,13 +36,15 @@
     <th scope="col" data-sortable="true">Status</th>
     <th scope="col" data-sortable="true">Running Since</th>
     <th scope="col">VMs</th>
-    <th scope="col">Details</th>
+    <th scope="col">timeslot reservation</th>
```
fyndalf: Please make this upper-case, like so:

```suggestion
    <th scope="col">Timeslot Reservation</th>
```

Thanks! :innocent:

PR #127, opened 2018-11-30
102/login/authorization redirect

janwestphal306:
If you are not logged in you will be redirected to the sign in page.

You might want to call `db:seed` to have a user in order to use the frontend.

Tests needed to be fixed to fit the authorization. This needs to be regarded for future tests.


app/controllers/landing_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
 class LandingController < ApplicationController
+  skip_before_action :authenticate_user!
```
lassesteffen: Pure rails magic again :+1:

spec/features/sessions/user_authorization_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'Authorization', type: :feature do
+  it 'redirects the user if not logged in' do
```
lassesteffen: Use context as describe here http://www.betterspecs.org/

The test should have two contexts 'if user is logged in' and 'if user is not logged in'. The 'it' should than only state what we expect e. g. 'it redirects the user', 'it shows a message', it should not explain the context.

spec/rails_helper.rb
```diff
@@ -70,7 +70,9 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
   config.include FeatureSpecHelper, type: :feature
+  # config.include FeatureSpecHelper, type: :request
```
lassesteffen: What are you doing here?

spec/requests/requests_spec.rb
```diff
@@ -3,10 +3,16 @@
...
+  before do
+    user = FactoryBot.create :user
+    sign_in user
+  end
+
   describe 'GET /requests' do
     it 'works! (now write some real specs)' do
-      get requests_path
-      expect(response).to have_http_status(200)
+      # get requests_path
```
lassesteffen: What is the reasoning for commenting out a test? Why not remove it, if it does not make sense?

config/locales/devise.en.yml
```diff
@@ -14,7 +14,7 @@ en:
       last_attempt: "You have one more attempt before your account is locked."
       not_found_in_database: "This user account does not exist."
       timeout: "Your session expired. Please sign in again to continue."
-      unauthenticated: "You need to sign in or sign up before continuing."
+      unauthenticated: "You need to sign in before continuing."
```
mlichtblau: Nice

spec/features/sessions/user_authorization_spec.rb
mlichtblau: Yes good catch!

spec/requests/requests_spec.rb
mlichtblau: It seems to be quite the dumb test...

> (now write some real tests)

what's that supposed to mean?

spec/controllers/requests_controller_spec.rb
```diff
@@ -26,6 +26,12 @@
 # `rails-controller-testing` gem.
 
 RSpec.describe RequestsController, type: :controller do
+  # Authenticate an user
+  before do
+    user = FactoryBot.create :user
+    sign_in user
```
mlichtblau: If you want to save a line you could write `sign_in FactoryBot.create(:user)` as you don't actually use the user variable anywhere else.

lassesteffen:
Seems good now, would be cool if you can update the branch and than merge this.

spec/features/sessions/user_authorization_spec.rb
janwestphal306: Thanks. It really looks better that way.

spec/requests/requests_spec.rb
janwestphal306: I did not see this in the commit. The test still exists but is now uncommented.

spec/controllers/requests_controller_spec.rb
janwestphal306: I implemented this your way. If some test need the user, it can always become a variable.

Issue #115, opened 2018-11-28
roles - user


isabel-amon:
**Beschreibung**
As a user of the tool I want to only see vms and their details if I an part of them, so that I don't have to worry about other people's things.
I want to be able to log into the tool as well as logout.
I also want to be able to see all profile pages, but only edit my own ssh-key.

(Every HPI-account is a user when it is logged in for the first time)

**Akzeptanzkriterien**
When a new account is logged in, it has the role of a user
A user can only see his/her own vms and their corresponding detail-pages
A user can see all profile pages
Only his/her own shh-key can be edited

**Priorität**

**Aufwandsschätzung**

skydivin4ng3l:
#99 is blocking the the adaption of the VM index page

isabel-amon:
the problem here is the second acceptance criterion, right? In case of this issue I see this functionality as critical for the purpose, so I wouldn't want to take it out...

lassesteffen:
Still blocked, same reason as in #221

skydivin4ng3l:
Komplexität?

Issue #109, opened 2018-11-28
roles - admin


isabel-amon:
**Beschreibung**
As an admin I want to have all rights a wimi has, and additionally I want to be able to:
- see all existing vms
- manage user rights (#110)
- see and manage vm-requests (#111)
- edit vm-templates (#98, should already be done)

in order to manage the tool efficiently.
The only thing I do not want to be able to do is see and change the ssh-key of a user that is not myself.

**Akzeptanzkriterien**
- [x] role of admin exists
- [x] stories #110 and #111 should be fulfilled
- [ ] only admin can add and edit vm-templates
- [x] the vm-page shows all existing vms
- [x] ssh-keys of other users are not visible or editable

**Priorität**

**Aufwandsschätzung**

**Tasks**
- [x] add admin to role enum on users
- [x] all employee restricted pages are also visible for admins
- [x] no filter from vm index page for admin
- [x] do #110 and #111

skydivin4ng3l:
kind of depends on #112

BjoernDaase:
@isabel-amon Please also add the ability to manage VM templates as an action an admin exclusively can do while WiMis (employees) and Users are not allowed to.
This is related to #98 which is finished and will hopefully merged today or tomorrow.

lassesteffen:
As @skydivin4ng3l said, this should only happen after #112 is done

skydivin4ng3l:
Komplexität

lassesteffen:
@isabel-amon, an employee can currently see all vms on the index page, is that the wanted behaviour?

isabel-amon:
@lassesteffen no, everyone should just see the vms that he is assigned to on the dashboard, except the admin, who sees all vms.

isabel-amon:
please don't forget about the vm-template editing.

lassesteffen:
Open Tasks
- [ ] hide edit vm templates for everyone but admins
- [ ] also filter vm for employees on vms/

PR #128, opened 2018-11-30
Add roles for User, Wimi and Admin

mlichtblau:
Users can now have different roles: User, Wimi and Admin.
References #109 #112 #115


.rubocop.yml
```diff
@@ -6,6 +6,7 @@ Metrics/BlockLength:
   - 'Rakefile'
   - '**/*.rake'
   - 'spec/**/*.rb'
+  - 'db/schema.rb'
```
lassesteffen: Thanks!

PR #135, opened 2018-12-03
73/oldies/integration of buttons

### LINKED ISSUE #73: Integration of buttons on the VM overview / detail page
As a user, I would like to be able to trigger different actions or reach subpages with different buttons on the VM overview page / detail page.

 - [x] There is a button on the VM overview page to create a new VM
    - This links to `/vms/new`
- [x] There is a button on the VM overview page to get an overview of all VM requests
    - This links to `/vms/requests`
- [x] There is a button on every VM in the overview page to start or shut down
    - Depending on the current start / stop status, only the appropriate one will be displayed
- [x] There are 6 energy management buttons on each VM detail page (see also vSphere, Start / Stop / Kill / Reset / shut down OS / restart OS)
    - Depending on the current status only suitable buttons are displayed
- [x] There is a button on every VM detail page to delete the VM.
    - This should currently be disabled because the deletion process is not yet defined
- [x] A click on the name of the server leads to the detail page (no button, but simple link)
- [x] VM boot time should only be displayed when powered on

Priority: medium

dey4ss:
resolve #73

### DIFF
--- a/app/controllers/vms_controller.rb
+++ b/app/controllers/vms_controller.rb
@@ -16,7 +16,7 @@ class VmsController < ApplicationController
 
   def destroy
     # params[:id] is actually the name of the vm, since the vsphere backend doesn't identify vms by IDs
-    VmApi.instance.delete_vm(params[:id]) if VmApi.instance.connected?
+    # VmApi.instance.delete_vm(params[:id]) if VmApi.instance.connected?
   end
 
   def create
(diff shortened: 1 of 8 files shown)


app/controllers/vms_controller.rb
```diff
@@ -16,7 +16,7 @@ def index
 
   def destroy
```
fewerner: destroy does nothing, why is this in here?

PR #138, opened 2018-12-04
99/scaffold - add table to add users in requests

Kenneth-Schroeder:
Related Issue: #99


Kenneth-Schroeder:
I don't think fixing codebeat it necessary here since it's just an "**info**-level-issue".

app/controllers/requests_controller.rb
```diff
@@ -100,13 +101,22 @@ def request_accept_button
 
   private
 
+  def save_sudo_rights(request)
+    sudo_users_for_request = request.users_assigned_to_requests.select { |uatq| request_params[:sudo_user_ids].include?(uatq.user_id.to_s) }
```
MaxSchneider1337: what is uatq? if its user assign to request then this should be called userToRequest connection, assignedRequest or assoctiation like in the next line

MaxSchneider1337: maby write the select block in multiple lines (this line seems way to long)

app/controllers/requests_controller.rb
```diff
@@ -100,13 +101,22 @@ def request_accept_button
...
+
+    sudo_users_for_request.each do |association|
+      association.sudo = true
+    end
```
MaxSchneider1337: the method is called save_sudo rights but the save is done elsewhere (line 42)

app/views/requests/show.html.erb
```diff
@@ -39,6 +40,30 @@
...
+  </p>
+  <% end %>
+  </ul>
+</p>
+
+<p>
+  <strong>Users with sudo rights:</strong>
+  <ul>
+    <% @request.users_assigned_to_requests.each do |userAtRequest| %>
+      <% if userAtRequest.sudo %>
```
MaxSchneider1337: maby make this a helper method or model method that returns all sudo users

db/migrate/20181209184959_create_users_assigned_to_requests.rb
```diff
@@ -0,0 +1,16 @@
...
+class CreateUsersAssignedToRequests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :users_assigned_to_requests do |t|
+      t.boolean 'sudo'
+      t.integer 'request_id'
+      t.integer 'user_id'
+      t.datetime 'created_at', null: false
+      t.datetime 'updated_at', null: false
+      t.index ['request_id'], name: 'index_users_assigned_to_requests_on_request_id'
+      t.index ['user_id'], name: 'index_users_assigned_to_requests_on_user_id'
```
MaxSchneider1337: are t.index ['request_id'] and t.index ['user_id'] really necessary you seem to never use them elsewhere?

Kenneth-Schroeder:
merged functionality, continued in #183

PR #126, opened 2018-11-30
Sprint 1

MkuuWaUjinga:
This PR contains all the user stories that were completed in the last sprint.
- There is a VM and a Server page + details page for each one of them.
- ...

PR #139, opened 2018-12-05
Sprint 1

marcfreiheit:
Second try for #126


MkuuWaUjinga:
@fyndalf could you have a look at the conflicting file `vmapi.rb`? I guess, it was just a renaming of variables to make conceptually clear that the hosts are actually clusters... 
`extract_clusters` seems to be an exact replica of `extract_hosts`. Can we delete the latter?

fyndalf:
> 
> 
> @fyndalf @dey4ss could you have a look at the conflicting file `vmapi.rb`? I guess, it was just a renaming of some variables to make conceptually clear that the hosts are actually clusters...
> `extract_clusters` seems to be an exact replica of `extract_hosts`. Can we delete the latter?

@MkuuWaUjinga You're correct, `extract_hosts` can be deleted as far as I can tell. I guess there was some funky whitespace stuff going on, so the conflict could not be properly resolved automatically. I'll just go ahead and resolve the conflict, and then this MR will be good to go :)

PR #156, opened 2018-12-06
33 HPI Login

jakobmache:
To enable HPI login, you need to provide the environment variables `client_id`, `client_secret` and `redirect_uri`. To get a client id and a client secret, visit `https://oidc.hpi.de`. The redirect uri for development should be `http://localhost:3000/users/auth/hpi/callback`. The host on the oidc site should be `http://localhost:3000` for development.


jakobmache:
We should think about a better way for providing a development setup than using the environment variables.

jakobmache:
In development, the authentication service will always log you in as a test user. For testing the hpi login, comment the line `OmniAuth.config.test_mode = true` out in `app/config/environments/development.rb`. 

You have to set the environment variables `client_id`, `client_secret` and `request_uri` on your machine. For this you have to register your OpenID Application at `oidc.hpi.de`. 
Then set the values, for example on the vagrant vm with `export client_id=your_client_id`, `export client_secret=your_client_secret` and `export redirect_uri=your_redirect_uri`.
The redirect uri should contain the url to the openid auth callback, `base_url/users/auth/hpi/callback`.

Gemfile.lock
```diff
@@ -426,6 +476,7 @@ DEPENDENCIES
   tzinfo-data
   uglifier (>= 1.3.0)
   web-console (>= 3.3.0)
+  with_advisory_lock
```
skydivin4ng3l: do not mind the position of this comment. What is the purpose of this gemfile.lock file?

.ruby-version
```diff
@@ -1 +1 @@
-2.5.0
\ No newline at end of file
+2.5.0
```
skydivin4ng3l: what happend here why is this changed?

README.md
```diff
@@ -88,4 +88,4 @@ Please keep in mind that this method may lead to a loss in performance, due to t
 * `bundle exec rails about` Show stats on current Rails installation, including version numbers
 
 ### Generating
-* `rails g migration DoSomething` Create migration _db/migrate/*_DoSomething.rb_.
+* `rails g migration DoSomething` Create migration _db/migrate/*_DoSomething.rb_.
```
skydivin4ng3l: what has been changed here? if there is no change should it really be updated?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    if @user.persisted?
+      sign_in_and_redirect @user, event: :authentication
+    else
+      flash[:danger] = 'Login failed'
+      redirect_to root_path
+    end
+  end
+
+  def failure
```
skydivin4ng3l: why do we not use this method in the else block above? is this even a method?

app/models/user.rb
```diff
@@ -24,15 +28,36 @@ def notify_slack(message)
...
+    end
+  end
+
   private
 
   def set_default_role
     self.role ||= :user
   end
+
+  def set_user_id
```
skydivin4ng3l: what about this user ids have to be > 4000 have i dream that or is this not the place to implement it?

app/views/devise/sessions/new.html.erb
```diff
@@ -39,7 +39,9 @@
           <% end -%>
           <div class="form-group actions">
             <%= f.submit "Log in", class: 'btn btn-block btn-primary float-right' %>
-            <button class='btn btn-block btn-secondary float-right' id="oauth">Sign in using OAuth</button>
+            <a class='btn btn-block btn-secondary float-right' href='<%= omniauth_authorize_path(:user, :hpi) %>'>
+              Sign in with HPI OpenID Connect
+            </a>
```
skydivin4ng3l: is the styling compliance with the style guides yet?

db/migrate/20181206152910_add_omniauth_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddOmniauthToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :provider, :string
+    add_column :users, :uid, :string
```
skydivin4ng3l: why is the user id a string here? or is uid not the user_id?

db/schema.rb
```diff
@@ -12,79 +10,84 @@
...
-    t.integer 'user_id'
-    t.string 'title'
-    t.string 'message'
-    t.boolean 'read'
-    t.datetime 'created_at', null: false
-    t.datetime 'updated_at', null: false
-    t.index ['user_id'], name: 'index_notifications_on_user_id'
+ActiveRecord::Schema.define(version: 2018_12_09_184959) do
+
+  create_table "notifications", force: :cascade do |t|
```
skydivin4ng3l: I have a feeling that we somehow should commit ourselves to ' or "?! is this always generated differently?

spec/features/sessions/user_sign_in_spec.rb
```diff
@@ -37,4 +37,38 @@
...
+                  first_name: 'Max',
+                  last_name: 'Mustermann',
+                  email: 'max.mustermann@student.hpi.de'
+              }
+          })
+      end
+
+      it 'performs a successful login' do
+        page.click_link 'Sign in with HPI OpenID Connect'
+        expect(page).to have_text('Logout')
```
skydivin4ng3l: Cool idea =)

spec/models/user_spec.rb
```diff
@@ -33,4 +33,68 @@
     user = FactoryBot.create :user
     expect(user).to be_truthy
   end
+
+  describe 'creating a user with openid connect' do
+    let(:auth) { double }
+    let(:info) { double }
```
skydivin4ng3l: does the  { double } here mean that there will be created any double type number?

spec/models/user_spec.rb
```diff
@@ -33,4 +33,68 @@
...
+        expect(user.last_name).to eq('Last')
+      end
+
+      it 'should set the users email' do
+        expect(user.email).to eq(mail)
+      end
+    end
+
+    context 'when the user already exists' do
+      let!(:existing_user) { FactoryBot.create :user, uid: 1, provider: 'testprovider', email: 'oldemail@mail.com' }
```
skydivin4ng3l: why do we use let! here with an !

app/models/user.rb
jakobmache: This method takes the maximum of the already existing user ids and increments it. If no maximum user id exists it it takes the configuration value `start_user_id`, which is currently set to 4000. So this is read from the config so we can change it easily later.

spec/models/user_spec.rb
jakobmache: `double` creates a mock object so we can stub method calls, see [here](https://relishapp.com/rspec/rspec-mocks/docs/basics/test-doubles)

spec/models/user_spec.rb
jakobmache: We use `let!` so the user gets created even if we not call the `existing_user` block. See[ here](https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let)

db/schema.rb
jakobmache: maybe we should discuss this with the other teams, so everybody can configure his rails the same way

app/views/devise/sessions/new.html.erb
jakobmache: i think so, it looks the same like the old button

app/controllers/users/omniauth_callbacks_controller.rb
jakobmache: it is a method which will be called when something in the omniauth login fails. So it does a semantically different thing (omniauth login process failed - something was not ok with openid or so vs persisting the user failed - the server provided malformed data or we did something wrong when reading an existing user)

README.md
jakobmache: i think i can remove it

spec/models/user_spec.rb
skydivin4ng3l: thx

PR #157, opened 2018-12-06
applied some style; should fix #59

### LINKED ISSUE #59: Übernahme des globalen Stylings für VM-Anfragen
**Beschreibung**
Ich als Nutzer möchte, dass die bisherigen Seiten in der neuen, globalen Optik verfügbar sind um mich auf diesen intuitiv bewegen zu können.

**Akzeptanzkriterien**
- [x] Alle bisher erstellten Seiten (/vms/request/*) haben den gemeinsamen Header und Footer aus #3 
- [x] Außerdem halten sie sich an den festgelegten Design-Standard aus #4

**Priorität**
mittel

**Aufwandsschätzung**

BraunTom:
fixes #59

### DIFF
--- a/app/views/requests/show.html.erb
+++ b/app/views/requests/show.html.erb
@@ -1,39 +1,45 @@
-<p id="notice"><%= notice %></p>
+<!-- <p id="notice"><%= notice %></p> -->
 
-<p>
-  <strong>Name:</strong>
-  <%= @request.name %>
-</p>
+<h1>Request Information</h1>
 
-<p>
-  <strong>Cpu cores:</strong>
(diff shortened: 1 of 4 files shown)


app/views/requests/show.html.erb
```diff
@@ -1,39 +1,45 @@
-<p id="notice"><%= notice %></p>
+<!-- <p id="notice"><%= notice %></p> -->
```
Kenneth-Schroeder: Why is this commented and not removed? :-)

BraunTom: I was not sure if someone wants to use it and only commented it

PR #160, opened 2018-12-07
Restrict access to request functionality

### LINKED ISSUE #112: roles - wimi
**Beschreibung**
As a wimi I want to be able to do everything a normal user can do, plus:
- request vms (#113 )

**Akzeptanzkriterien**
- [x] Role of wimi exists
- [ ] story #113

**Priorität**

**Aufwandsschätzung**

**Tasks**
- [x] add wimi to role enum
- [ ] whitelist vm request page
- [ ] do #113

mlichtblau:
Restrict access to all Request functionality to wimis and admins.

fixes #112

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,10 +14,10 @@ class ApplicationController < ActionController::Base
   end
 
   def authenticate_wimi
-    redirect_to root_path, alert: 'You don\'t have the necessary rights to perform this operation' unless current_user && (current_user.admin? || current_user.wimi?)
+    redirect_to root_path, alert: I18n.t('authorization.unauthorized') unless current_user && (current_user.admin? || current_user.wimi?)
   end
 
   def authenticate_admin
-    redirect_to root_path, alert: 'You don\'t have the necessary rights to perform this operation' unless current_user && current_user.admin?
+    redirect_to root_path, alert: I18n.t('authorization.unauthorized') unless current_user && current_user.admin?
   end
 end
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -12,4 +12,12 @@ def after_sign_in_path_for(_resource)
   def error_render_method
     render(template: 'errors/timeout', status: 408) && return
   end
+
+  def authenticate_wimi
+    redirect_to root_path, alert: 'You don\'t have the necessary rights to perform this operation' unless current_user && (current_user.admin? || current_user.wimi?)
```
skydivin4ng3l: is the root_path here always the best option? can it happen that a user is on a specific site and will be redirected to the root just by clicking on links he is not supposed to and confused. Is there a notification we could use to inform that he has no access, would it be necessary?

westphal-jan (approved):

app/views/vms/index.html.erb
```diff
@@ -1,3 +1,9 @@
+<% flash.each do |key, value| %>
+  <div class="alert alert-danger">
+    <%= content_tag :div, value %>
+  </div>
+<% end %>
+
```
skydivin4ng3l: what does this do? which alert?

skydivin4ng3l (approved):

app/controllers/application_controller.rb
mlichtblau: The notification is already there. We probably could remember from which path the user is coming from and then redirect back to it, but I think that's out of scope for this user story

PR #181, opened 2018-12-10
65/frontend/dashboard notifications

henleo:
dependencies: #103 #105 
do not merge before the dependencies are merged
apart from that, everything should be done here


NiklasKoehnecke:
I did not find a way to create a notification so i could not look at the layout. The code seems to be fine tho.

henleo:
> The Dashboard and VM's links in the header are broken, the margin around the page is back again. Not sure whose fault it is but it should be avoided to merge that back into dev.

This should be resolved by merging #105 
This branch is based on #63 so those links simply are not implemented here

PR #188, opened 2018-12-13
95/scaffold/add ports to requests

regnujAx:
about issue #95 

Adds a port and a name from which the vm can be reached from outside.


app/views/requests/index.html.erb
```diff
@@ -6,10 +6,12 @@
...
     <tr>
       <th>Name</th>
-      <th>Cpu-Cores</th>
-      <th>MB RAM</th>
-      <th>MB Storage</th>
+      <th style="width: 80px">Cpu-Cores</th>
+      <th style="width: 80px">MB RAM</th>
+      <th style="width: 90px">MB Storage</th>
       <th>Operating system</th>
+      <th style="width: 80px">Port</th>
```
Kenneth-Schroeder: Why did you add a width for those columns? :-)

Kenneth-Schroeder:
Reminder for next time: Please remember to add labels to your PR (WIP or Review required)

Kenneth-Schroeder:
@BjoernDaase What is the difference between the Request name and the ("reachability") name of the VM? Do we need a separate name for the requests at all? It seems like a single field "VM Name" would be enough.

BjoernDaase:
@Kenneth-Schroeder The difference is basically the following:
The VM name (which we already specify in the request, calling it only "Name") is the internal name of the VM which is used for all management activities of the VM.
What you called the "reachability name" is the name of the application running on the VM which should be accessible from outside.
So the first one is our internal name (e.g. "Special VM for seminar XYZ") and the second one is the applications name they in the build in this seminar and want to share it with everybody out there which then could be "ABC". So that must have nothing to do with the VMs name.
Does that make it clear?

Kenneth-Schroeder:
@regnujAx Would you mind changing "reachability name" to "application name" and "name" to "VM name"?
That would be great. (=

app/views/requests/index.html.erb
regnujAx: Because of the layout. It's look better with the width ;)
But I can delete it if you think it is to "hardcoded".

Kenneth-Schroeder: Maybe we could/should use something like this: https://stackoverflow.com/questions/18503849/set-min-width-in-html-tables-td

regnujAx:
@Kenneth-Schroeder I'll do it with the names in the next commit ;)
Btw: I named it "reachability name" because I was unimaginative and I wasn't sure how I can reduce "name of the application running on the VM which should be accessible from outside" ^^' 
Thanks for the new name :)

config/database.yml
```diff
@@ -18,7 +18,7 @@ development:
 # Do not set this db to the same as development or production database.
 test:
   <<: *default
-  database: tmp/test.sqlite3
+  database: db/test.sqlite3
```
Kenneth-Schroeder: What does this change do? Is it important? :-)

regnujAx: It is only the folder where test.sqlite3 is saved. The folder is not important but I change it to db so every .sqlite3 file is in the same folder ;)

BjoernDaase:
@regnujAx @Kenneth-Schroeder I think you might have bricked something...
screenshot from 2018-12-16 11-50-27

PR #189, opened 2018-12-13
146/login/Successful-login-logout-messages-green

skydivin4ng3l:
add refactored and styled flash messages to landing vms and sign_in page, remove duplicate root \vms root is still landing#index, changed logout flash message from danger to success, added hook for different sign_out path. after logout there will be a green flash message "successfully logged out"

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,11 @@ class ApplicationController < ActionController::Base
     vms_path
   end
 
+  # Overwriting the sign_out redirect path method
+  def after_sign_out_path_for(_resource)
+    root_path
+  end
+
   rescue_from Net::OpenTimeout, Errno::ENETUNREACH, Errno::EHOSTUNREACH, with: :error_render_method
 
   def error_render_method
--- a/app/views/application/_navbar.html.erb
+++ b/app/views/application/_navbar.html.erb
@@ -20,7 +20,7 @@
       </li>
       <% if user_signed_in? %>
         <li>
-          <%= link_to('Logout', destroy_user_session_path, method: :delete, class: "nav-link text-danger") %>
+          <%= link_to('Logout', destroy_user_session_path, method: :delete, class: "nav-link text-success") %>
         </li>
       <% else %>
         <li>
--- a/app/views/landing/index.html.erb
+++ b/app/views/landing/index.html.erb
@@ -1,3 +1,5 @@
+<%= render('layouts/flash_message') %>
+
 <h1>Welcome to HART<br/>
 <small class="text-muted">Landing#index</small></h1>
 <p>Find me in app/views/landing/index.html.erb</p>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,7 +6,7 @@ Rails.application.routes.draw do
   patch 'requests/request_accept_button', to: 'requests#request_accept_button', as: 'request_accept_button'
   resources :requests, path: '/vms/requests'
 
-  root to: redirect('/vms')
+  # root to: redirect('/vms')
 
   get '/servers/:id' => 'servers#show', constraints: { id: /.*/ }
 


skydivin4ng3l:
do not understand whats failing locally all test are green and manually also the test cases are working.

I think this could be responsible: **E, [2018-12-13T22:41:18.394269 #11871] ERROR -- omniauth: (hpi) Authentication failure! invalid_credentials encountered.** or it is to be expected case we test for an error? no clue

spec/features/sessions/user_sign_in_spec.rb
```diff
@@ -56,6 +67,9 @@
       it 'performs a successful login' do
         page.click_link 'Sign in with HPI OpenID Connect'
         expect(page).to have_text('Logout')
+        #expect(page).to have_text 'Signed in successfully.'
+        expect(page).to have_text 'Successfully authenticated'
```
jakobmache: pls move this to a separate test which tests if there is a flash message

skydivin4ng3l: but then i would have to write the exact same test?

skydivin4ng3l: I see will separate in the same file

app/controllers/application_controller.rb
```diff
@@ -3,10 +3,17 @@
 class ApplicationController < ActionController::Base
   before_action :authenticate_user!
 
+  add_flash_types :success, :danger
```
lassesteffen: is this still necessary?

app/views/errors/timeout.html.erb
```diff
@@ -1,3 +1,4 @@
+
```
lassesteffen: Can you remove the new line again?

app/views/layouts/_flash_message.html.erb
```diff
@@ -0,0 +1,8 @@
+<div>
+  <% flash.each do |key, value| %>
+    <div class="alert alert-<%= translate_flash_key(key) %>">
+      <a href="#" data-dismiss="alert" class="close">×</a>
+      <%= content_tag :div, value %>
+    </div>
+  <% end %>
+</div>
```
lassesteffen: New line at the end of the file

spec/features/sessions/user_sign_in_spec.rb
```diff
@@ -23,40 +24,62 @@
     expect(page).to have_link 'Sign in with HPI OpenID Connect'
   end
 
-  it 'shows an error message if email is not found' do
-    page.fill_in 'user[email]', with: 'login@email.com'
-    page.fill_in 'user[password]', with: 'secret_password'
-    page.find('input[type=submit]').click
-    expect(page).to have_text 'This user account does not exist.'
-  end
+  context 'with classic login' do
```
lassesteffen: Better contexts plsss

spec/features/sessions/user_sign_in_spec.rb
```diff
@@ -69,6 +92,12 @@
         expect(page).to_not have_text('Logout')
         expect(page).to have_text('Login')
       end
+
+      it 'shows Danger Flash Message after not successful login' do
```
lassesteffen: context

app/controllers/application_controller.rb
skydivin4ng3l: i would imagine that it is still necessary cause i think devise might intervene and change the type and we then again convert it. But this is just a guess

skydivin4ng3l: okay this is not from devise but from action pack gem and i would recommend using it to be allowed to use more than just alert and notice to be send.

skydivin4ng3l: # Creates new flash types. You can pass as many types as you want to create
      # flash types other than the default <tt>alert</tt> and <tt>notice</tt> in
      # your controllers and views. For instance:
      #
      #   # in application_controller.rb
      #   class ApplicationController < ActionController::Base
      #     add_flash_types :warning
      #   end
      #
      #   # in your controller
      #   redirect_to user_path(@user), warning: "Incomplete profile"
      #
      #   # in your view
      #   <%= warning %>
      #
      # This method will automatically define a new method for each of the given
      # names, and it will be available in your views.

PR #197, opened 2018-12-14
Move "add to slack" button to user page

MrModder:


MrModder:
This Pull request now also changes the request description text a bit as requested by @BjoernDaase today.

app/models/request.rb
```diff
@@ -12,17 +12,21 @@ class Request < ApplicationRecord
...
     description
   end
 
   def accept!
     self.status = 'accepted'
   end
+
+  private
+
+  def get_url(host_name)
```
mlichtblau: Is `get_url` a good name? Maybe `url` is sufficient?

MrModder: Thank you for the review.
Fair point, the get_ does not really add value, I changed the name as suggested.

PR #200, opened 2018-12-14
110 Better manage user roles

### LINKED ISSUE #110: manage user roles
**Beschreibung**
As an admin I want to be able to manage the rights of the users of our tool, so that everyone can use the functionalities  he/she needs. On a all-users page I want to see a grouped toggle button (see screenshot below) with the options of user, employee and admin. The current role of the user should be pressed down so I can see it clearly. 
When I click on a role the user not currently has, I want his or her role changed accordingly.

screen shot 2018-12-19 at 2 27 16 pm

I only want to be able to change my own role, if there is still one other admin left after I do so, in order to avoid there being no admin left. In this case the other options next to my name in my grouped button should be greyed out and not clickable.

**Akzeptanzkriterien**
- [x] A grouped button exists on the all-users-page, displaying the users current role
- [x] This button is only visible for admins
- [x] Pressing another role changes the role of the according user
- [x] After the change happened the user gets access to the tool according to his new role
- [x] An admin can only change his/her own role if ensured that there is at least one other admin left
- [x] If no other admin would be left, the last admins other role buttons are not clickable

**Priorität**

**Aufwandsschätzung**

lassesteffen:
Fixes #110


app/helpers/users_helper.rb
```diff
@@ -1,4 +1,17 @@
 # frozen_string_literal: true
 
 module UsersHelper
+  def last_admin?(user)
+    user.admin? &&
+      current_user?(user) &&
```
mlichtblau: Why do you need that? Does the function `last_admin?` really want to check if the given user is the current user?

app/views/users/index.html.erb
```diff
@@ -2,15 +2,40 @@
...
   <tbody>
     <% @users.each do |user| %>
       <tr>
         <td><%= link_to user.name, user_path(user) %></td>
         <td><%= user.email %></td>
+        <% if current_user.admin? %>
+          <td>
+            <div class="btn-group" role="group">
+              <% User.roles.map(&:first).each do |role| %>
+                <% if last_admin?(user) %>
```
mlichtblau: I think you can write `last_admin?(user) && current_user?(user)` or define a new function to avoid having `current_user?` inside the `last_admin` function.

spec/factories/users.rb
```diff
@@ -7,10 +7,18 @@
...
     ssh_key { SSHKey.generate.ssh_public_key }
   end
+
+  factory :admin, parent: :user do
+    role { :admin }
+  end
+
+  factory :employee, parent: :user do
+    role { :employee }
+  end
```
mlichtblau: nice, nice!

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class UsersController < ApplicationController
+  include UsersHelper
+
   before_action :authenticate_current_user, only: [:edit]
```
mlichtblau: I think you should make sure that the user is an admin to update a role.
`before_action :authenticate_admin, only: [:update_role]`

app/helpers/users_helper.rb
lassesteffen: That was the plan. The requirement is that an admin is the only one that can change roles and that he can not change his own role if the is the last admin left. I will split it up :+1:

lassesteffen:
@mlichtblau I implemented the suggestions, would be nice if I could get another review

Issue #155, opened 2018-12-06
Redirect to the dashboard after requesting a VM


BjoernDaase:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

As an employee I want to be redirected to my dashboard after requesting a VM because the current page (the accept & reject page for am VM) is a page I am not entitled to see.

**Acceptance Criteria:**

- [ ] When clicking the “Request VM” button on the request VM page I want to be redirected to my dashboard when I’m not an admin
- [ ] When clicking the “Request VM” button on the request VM page and I’m an admin I want to keep directed to the Accept & Reject page

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

BjoernDaase:
Looks good to me!

PR #202, opened 2018-12-15
155/scaffold custom redirection after vm request creation

Kenneth-Schroeder:
corresponding issue: #155

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -31,7 +31,11 @@ class RequestsController < ApplicationController
 
   def successfully_saved(format, request)
     notify_users("New VM request:\n" + request.description_text)
-    format.html { redirect_to @request, notice: 'Request was successfully created.' }
+    if current_user.role == 'admin'
+      format.html { redirect_to @request, notice: 'Request was successfully created.' }
+    else
+      format.html { redirect_to requests_url, notice: 'Request was successfully created.' }
+    end
     format.json { render :show, status: :created, location: request }
   end
 
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -29,13 +29,17 @@ class RequestsController < ApplicationController
     end
   end
 
-  def successfully_saved(format, request)
-    notify_users("New VM request:\n" + request.description_text)
-    if current_user.role == 'admin'
+  def redirect_according_to_role(format, role)
+    if role == 'admin'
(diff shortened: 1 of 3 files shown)


spec/controllers/requests_controller_spec.rb
```diff
@@ -99,9 +99,9 @@
         end.to change(Request, :count).by(1)
       end
 
-      it 'redirects to the created request' do
+      it 'redirects to the request overview if user is not an admin' do
         post :create, params: { request: valid_attributes }
-        expect(response).to redirect_to(Request.last)
+        expect(response).to redirect_to(requests_url)
       end
```
Kenneth-Schroeder: @MrModder is there an elegant way to create a second test, that tests the redirection for an admin user?
I am not sure how to do this, since for this file the signed in user seems to always be wimi.

app/controllers/requests_controller.rb
```diff
@@ -31,7 +31,11 @@ def notify_users(message)
 
   def successfully_saved(format, request)
     notify_users("New VM request:\n" + request.description_text)
-    format.html { redirect_to @request, notice: 'Request was successfully created.' }
+    if current_user.role == 'admin'
+      format.html { redirect_to @request, notice: 'Request was successfully created.' }
+    else
+      format.html { redirect_to requests_url, notice: 'Request was successfully created.' }
```
MrModder: According to #155 the user should be redirected to the dashboard, not the requests overview

spec/controllers/requests_controller_spec.rb
MrModder: You somehow have to create an admin user (should be possible by adapting the user factory code)
Then you can just use:
``` ruby
user = FactoryBot create :admin
sign_in(user) #this will sign in the user via devise, do this before you request anything
```
You can take a look at ```slack_controller_spec.rb``` the ```mock_sign_in``` function does this with a standard user.

app/views/requests/index.html.erb
```diff
@@ -1,5 +1,3 @@
-<p id="notice"><%= notice %></p>
```
MrModder: Why is the notice removed here?
This change seems to remove value without a reason for it, especially as you are still adding a notice in the requests controller.

Kenneth-Schroeder: I removed it because I received double notices when I was falsely redirecting to this page. I will add it again.

Issue #178, opened 2018-12-07
Warning Pop-Ups


arne-z:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

As a user
I want to have a warning pop up when I am about to do something irreversible or potentially dangerous
in order to not accidentally click "delete" on something important.

**Acceptance Criteria:**

- [ ] A Template for Warnings exists and is made available to all teams to implement when needed.
- [ ] Warning Popups show up on the following occasions:
- When hitting the "Shut Down" and "Reset" buttons on a VM
- When archiving a VM
- Remove warning on /notifications page.

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

hkohlsaat:
Template for warnings would be over-engineering since the only thing that is required to warn about a critical action is appending `data: { confirm: 'Are you sure?' }` to the `link_to` or `button_to` helper calls.

Securing "Archive" on VM cannot be done since there aren't any "Archive" buttons, as far as I can see.

PR #209, opened 2018-12-17
Confirmation dialogs for potentially dangerous actions

hkohlsaat:


hkohlsaat:
@phSch08 I think the code is ready to be reviewed by you. For information in the task, I refer you to the [issue page](https://github.com/hpi-swt2/vm-portal/issues/178) and the comments there. The only check that was not successful is CodeFactor, which complaints about an issue that is non existent, if you look into the source files of the commits. I think the rest should be okay. Thank you for reviewing :)

PR #210, opened 2018-12-17
186 Remove 'notification removed' message

luisebenkert:


app/controllers/notifications_controller.rb
```diff
@@ -51,14 +51,6 @@ def mark_as_read
     end
   end
 
-  def any?
```
henleo: y u do dad???

PR #213, opened 2018-12-18
65/vertical/dashboard notifications

henleo:


henleo:
@NiklasKoehnecke 
test added, philipp notified
the link to the indexpage is in this particular user story so i thought i'd add it for the other elements as well in order to be consistent

PR #218, opened 2018-12-22
try to fix Safari hart-dev error

Kenneth-Schroeder:


Kenneth-Schroeder:
screenshot 2018-12-22 at 19 09 30

This PR resolved the exception but signing in still doesn't work. I will try further with another PR.

PR #219, opened 2018-12-22
217/scaffold/fix safari hart dev error ---- 2

Kenneth-Schroeder:


Kenneth-Schroeder:
these changes had the same effect as in #218

PR #226, opened 2019-01-07
149/scaffold/archive vms

MaxSchneider1337:


app/controllers/vms_controller.rb
```diff
@@ -24,36 +28,51 @@ def new
...
+    end
+    set_pending_archivation(@vm)
+
+    redirect_to controller: :vms, action: 'show', id: @vm.name
+  end
+
+  def change_power_state
+    @vm = VmApi.instance.get_vm_info(params[:id])
     VmApi.instance.change_power_state(@vm[:name], @vm[:state])
     redirect_to root_path
```
MaxSchneider1337: shouldn't this redirect to then vms show site

Issue #97, opened 2018-11-23
creating puppet scripts


arne-z:
**Beschreibung**
As an admin I want to receive a puppet-script after a VM request has been approved, in which I can find important information to my VM, so that I can use it to create accounts. I'd like to see a button that allows me to see the puppet-file in raw text, like the github "raw" button. This button should be found on the detail page for the approved request. 

Beispiel eines Puppet Scripts:
```
class node-vm-username-operatingsystem {
    $admins = ["FirstName.LastName", "FirstName2.LastName2"]  
    $users = ["FirstName3.LastName3." , ... ]  

    realize(Accounts::Virtual[$admins], Accounts::Sudoroot[$admins])  
    realize(Accounts::Virtual[$users])  
}  
All names are based on HPI-Active Directory accounts, 
therefore they can include numbers if there are two or more students with the same names.
```

**Akzeptanzkriterien**
- [ ] After a VM request has been accepted, a button should be visible on the VMs request page, that leads to its puppet file in plain text.

**Priorität**
High

**Aufwandsschätzung**

skydivin4ng3l:
references #100

skydivin4ng3l:
discuss with @hpi-swt2/bd-18 about how users get priviliges in the created VM, accessing User DB for instance

skydivin4ng3l:
discuss with @hpi-swt2/bd-18 possible contents of puppet files

skydivin4ng3l:
startet a discussion with @BjoernDaase about this.

PR #234, opened 2019-01-10
100/vertical/git puppet scripts

### LINKED ISSUE #100: Puppet-scripts pushed automatically to git
**Beschreibung**
As a user I want generated puppet scripts to be automatically pushed into a git repository, so I don't have to do it manually everytime.

(Jedes mal wenn ein neues Puppet-Script erstellt wird, soll dieses direkt in ein Git-Repo gepushed werden, in dem derzeit bereits alle solchen Scripte gesammelt werden. Da wir derzeit noch keinen Zugang zu diesem Repo haben, soll diese Funktionalität anhand von einem Dummy-Repo demonstriert werden, mit der Möglichkeit nachträglich das Repo in das gepushed wird einfach zu verändern.)

**Akzeptanzkriterien**
- [ ] After a script has been created it is automatically pushed into a git repo
- [ ] The repository being pushed to can be changed easily
 - [ ] Es kann ein SSH key für das Repo hinterlegt werden, damit kein Username und Password für das Git-Repo erforderlich sind

**Priorität**
High

**Aufwandsschätzung**

mewitte:
Closes #100.
This pushes created puppet scripts (e.g. #97) to a private git repository. It uses git credentials and config for the currently logged in user.


app/controllers/requests_controller.rb
```diff
@@ -134,4 +135,79 @@ def request_params
     params.require(:request).permit(:name, :cpu_cores, :ram_mb, :storage_mb, :operating_system,
                                     :port, :application_name, :comment, :rejection_information, :status, user_ids: [], sudo_user_ids: [])
   end
+
+  # TODO get credentials and config for currently logged in user
```
NiklasKoehnecke: When will this be done?

app/controllers/requests_controller.rb
```diff
@@ -134,4 +135,79 @@ def request_params
...
+  # Clones and configures a git repository on dir.
+  def setup_git(dir)
+    # Dir.mkdir(dir) unless File.exists?(dir)
+    uri = "git@github.com:mewitte/PuppetFiles.git"
```
NiklasKoehnecke: Is this "easy to change" for the po? if yes its ok, otherwise only devs can change it.

app/controllers/requests_controller.rb
```diff
@@ -134,4 +135,79 @@ def request_params
...
+    write_node_script(dir, request)
+    write_name_script
+    change_init_script
+
+    if g.status.untracked.length + g.status.changed.length == 0
+      g.commit_all("Add node " + request.name)
+      g.push
+    end
+
+    FileUtils.rm_rf(dir)
```
NiklasKoehnecke: cloning and deleting might be a lot of work later for a single script if the repository becomes large

app/controllers/requests_controller.rb
arne-z: The canonical way of solving this would be a .env file, See https://rubygems.org/gems/dotenv/versions/2.1.0?locale=de

app/controllers/requests_controller.rb
arne-z: Why is it necessary to delete the repo afterward? Would it add a lot of extra work to keep it around?
Performance is honestly not an issue for this application, because VMs are likely created at a very low rate.

app/controllers/requests_controller.rb
arne-z: I will talk to the customer in 20 minutes about getting access to the repo, hopefully, the credentials will be available then. Please do not hard code in the credentials though, this is a public repo.
Use a .env file or a secret for this (see below).

Adrian-St: Credentials will be in an env file. I don't think you need the current_user

app/controllers/requests_controller.rb
Adrian-St: No, but you need to talk to the customer it's okay to keep the repository.

lib/puppetscript.rb
```diff
@@ -0,0 +1,43 @@
...
+
+              realize(Accounts::Virtual[$admins], Accounts::Sudoroot[$admins])
+              realize(Accounts::Virtual[$users])
+      }
+    NODE_SCRIPT
+  end
+
+  def self.generic_name_script
+    <<~NAME_SCRIPT
+      node \'vm-%s\'{
```
NiklasKoehnecke: see above

spec/models/request_spec.rb
```diff
@@ -155,4 +155,142 @@
...
+
+  describe 'puppet script helper methods' do
+    before do
+      @request = FactoryBot.create(:request_with_users)
+    end
+
+    it 'returns correct declaration script for a given request' do
+      script = @request.generate_puppet_name_script
+      expected_string = <<~NAME_SCRIPT
+        node \'vm-MyVM\'{
```
NiklasKoehnecke: see above

spec/models/request_spec.rb
```diff
@@ -155,4 +155,142 @@
...
+            }
+        }
+      NAME_SCRIPT
+      expect(script).to eq(expected_string)
+    end
+
+    it 'returns correct initialization script for a given request' do
+      script = @request.generate_puppet_node_script
+      expected_string = <<~NODE_SCRIPT
+        class node_vm-MyVM {
```
NiklasKoehnecke: see above

PR #235, opened 2019-01-10
Show only vms associated with the current user

mlichtblau:
If the current user is a simple user only show his vms on the vm index page.

Also add associations between vms and requests to determine which users belongs to which request.

### DIFF
--- a/app/api/vmapi.rb
+++ b/app/api/vmapi.rb
@@ -24,7 +24,8 @@ class VmApi
       { name: vm.name,
         state: (vm.runtime.powerState == 'poweredOn'),
         boot_time: vm.runtime.bootTime,
-        vmwaretools: (vm.guest.toolsStatus != 'toolsNotInstalled') }
+        vmwaretools: (vm.guest.toolsStatus != 'toolsNotInstalled'),
+        request: Request.find_by_name(vm.name) }
     end
   end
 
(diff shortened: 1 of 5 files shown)


app/api/vmapi.rb
```diff
@@ -71,7 +72,9 @@ def get_vm(name)
         host: vm.summary.runtime.host.name,
         guestHeartbeatStatus: vm.guestHeartbeatStatus,
         summary: vm.summary,
-        vmwaretools: (vm.guest.toolsStatus != 'toolsNotInstalled') }
+        vmwaretools: (vm.guest.toolsStatus != 'toolsNotInstalled'),
+        request: Request.find_by_name(vm.name)
```
MrModder: Simply finding the Request by name is not enough!
The Request must also be accepted, otherwise the Request could not have created the VM and the two are not associated!
It must also be ensured that no two Requests that are accepted have the same name, this is probably not enforced at the moment.

MaxSchneider1337: We added vm_users(vm). It does not address all of these problems, but we will probably refactor the whole vmapi and fix the rest.

MaxSchneider1337: we also had to rename the vmapi methods that only retrun a hash and not the real vm object. we used "_info" for the f.e. all_vm_infos.

PR #236, opened 2019-01-10
UI Refactoring

luisebenkert:


BjoernDaase:
First of all I have to say that your changes look really good. I just found a few things that were already done on dev (after you branched this one) which aren't in your branch. Could it be that they are a bit out of sync (I especially saw the users, hosts, servers and projects page, the colors of the user roles and the user selection on the requests page)?

And a tiny "feature request": Could you place the bell and the little man right to it in the header on the same height? There seems to be a tiny margin.

app/controllers/notifications_controller.rb
```diff
@@ -51,14 +51,6 @@ def mark_as_read
...
   end
 
-  def any?
-    has_unread_notifications = Notification.where(user: current_user, read: false).any?
-    respond_to do |format|
-      msg = { has_unread_notifications: has_unread_notifications }
-      format.json { render json: msg }
-    end
-  end
-
```
Adrian-St: Why did you delete that?

PR #247, opened 2019-01-11
If current user has role user show only vms associated to user

### LINKED ISSUE #115: roles - user

mlichtblau:
Fixes #115

### DIFF
--- a/app/controllers/vms_controller.rb
+++ b/app/controllers/vms_controller.rb
@@ -9,7 +9,11 @@ class VmsController < ApplicationController
   before_action :authenticate_admin, only: %i[archive_vm]
 
   def index
-    @vms = filter VmApi.instance.all_vm_infos
+    if current_user.user?
+      @vms = filter current_user.vms
+    else
+      @vms = filter VmApi.instance.all_vm_infos
+    end
     @archived_vms = all_archived_vms
     @parameters = determine_params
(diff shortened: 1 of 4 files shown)


app/controllers/vms_controller.rb
```diff
@@ -31,7 +35,8 @@ def new
 
   def show
     @vm = VmApi.instance.get_vm_info(params[:id])
-    render(template: 'errors/not_found', status: :not_found) if @vm.nil?
+    return render(template: 'errors/not_found', status: :not_found) if @vm.nil?
+    return redirect_to vms_path unless !current_user.user? || current_user.vms.include?(@vm)
```
lassesteffen: Can you refactor this to a before_action?

lassesteffen: You can also use this for edit...

PR #248, opened 2019-01-11
Fix/improve vmapi

MrModder:
This pull requests attempts to provide a new API for interaction with vSphere, as the current one has turned into kind of a mess.
This pull request does not aim to refactor the current uses of the old API, this shall be done in another pull request, after this one is merged.


MrModder:
@mlichtblau 
Okay, thanks, we will then add your changes to the new API, as we want it to be completely capable of replacing the old one.
The PR is still WIP as we need to test it first, at it is not tested enough.
We will tell you when we remove the WIP label, so you can review the finished PR.

Issue #222, opened 2019-01-06
Remove VM creation indirection


BjoernDaase:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**
As an admin I want to need one click only to accept a VM request and create the corresponding VM in order to save my time. Therefore, the fields on the VM request acceptance page have to be editable and the "Accept & Create" (former "Accept & Continue") button will already create the VM.
 
![no_indirection](https://user-images.githubusercontent.com/24682367/50729069-bfc6ec80-1133-11e9-84c9-c612549729d2.png)

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

**Acceptance Criteria:**

- [ ] The "Accept & Continue" button now says "Accept & Create"
- [ ] When clicking the "Accept & Create" button the VM will be created directly
- [ ] On the accept-request page all fields are editable (for the admin)
- [ ] The direct VM creation page is completely removed

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

BjoernDaase:
I will update this one later today because we have a major update of the functionality.

PR #253, opened 2019-01-12
222/scaffold/revision_of_vm_creation

regnujAx:
about issue #222


app/views/requests/show.html.erb
```diff
@@ -3,75 +3,133 @@
 <table class="table table-fixed-columns table-striped">
   <tbody>
     <tr>
-      <th>VM Name:</td>
-      <td><%= @request.name %></td>
+      <td><strong>VM Name:</strong></td>
```
MaxSchneider1337: we should probaply use the form here or make 2 seperate files (one for the admins and one for the users etc.)

config/routes.rb
```diff
@@ -26,6 +26,8 @@
   get 'slack/new' => 'slack#new', as: :new_slack
   get 'slack/auth' => 'slack#update', as: :update_slack
 
+  get '/requests/:id/recap', to: 'requests#recap'
```
MaxSchneider1337: nice nameing :-D

app/controllers/requests_controller.rb
```diff
@@ -79,10 +82,20 @@ def destroy
     end
   end
 
+  def recap
+    @request = Request.find(params[:id])
+    @vm = VSphere::VirtualMachine.find_by_name(@request.name)
+  end
+
   def request_change_state
     if @request.update(request_params)
```
MaxSchneider1337: we still have to assign the sudo users (we should fix that in update too)

app/api/v_sphere/virtual_machine.rb
```diff
@@ -134,6 +134,10 @@ def ==(other)
       equal? other
     end
 
+    def macs
```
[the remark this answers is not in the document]
BjoernDaase: @MaxSchneider1337 It might have (@regnujAx and I discussed this yesterday).
I will ask the customer on Wednesday and then we will see whether we have to change something.

config/routes.rb
MaxSchneider1337: But shouldn't this be vms/requests/:id/recap ?

MaxSchneider1337: `patch 'requests/request_change_state', to: 'requests#request_change_state', as: 'request_change_state'` should also have a vms/ at the beginning I think (better ask @BjoernDaase )

app/views/requests/show.html.erb
```diff
@@ -88,7 +146,7 @@
     <%= form_with(model: @request, url:{action: 'request_change_state', id: @request}, local: true) do |form| %>
```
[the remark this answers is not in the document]
regnujAx: @MaxSchneider1337 
I have no error on show page.

app/views/requests/show.html.erb
```diff
@@ -57,21 +93,39 @@
...
 </p>
 
 <p>
   <strong>Users:</strong>
   <ul>
-    <% @request.non_sudo_user_assignments.each do |assignment| %>
-        <p>
-          <li><%= assignment.user.email %></li>
-        </p>
+    <% if current_user.admin? && @request.pending? %>
```
MaxSchneider1337: I also think wwe should completly rework this part, and use a accept button and not a seperate form (I dont know if this works with the controller emthod only)

app/views/requests/show.html.erb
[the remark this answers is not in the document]
MaxSchneider1337: the error appears when the request is accepted

app/controllers/requests_controller.rb
```diff
@@ -79,17 +79,17 @@ def destroy
 
   def recap
     @request = Request.find(params[:id])
+    return unless @request
+
     @vm = VSphere::VirtualMachine.find_by_name(@request.name)
     @request.assign_sudo_users(request_params[:sudo_user_ids][1..-1])
```
MaxSchneider1337: why do we assign the sudo users here. I think thsi should be done in theb change state, since this is where the rest of the request is updated

app/api/v_sphere/virtual_machine.rb
```diff
@@ -146,6 +151,19 @@ def set_revived
 
     # Config methods
     # All the properties that HART saves internally
+    def config
+      @config ||= VirtualMachineConfig.find_by_name name
```
MaxSchneider1337: all of the config methods are so clean I love it

PR #257, opened 2019-01-14
Use secrets.yml instead of environment variables

jakobmache:
We need to use secrets.yml instead of environment variables on production server. 
To use hpi login now:
Create file `secrets.yml` in `config` directory.
Fill it with the following content:
```
development:
    client_id: 'MY_CLIENT_ID'
    client_secret: 'MY_CLIENT_SECRET'
    redirect_uri: 'MY_REDIRECT_URI'
```
(Replace `development` with the environment you want to use for login, e.g. `production`)


jakobmache:
For setting this up: 
Run `rails credentials:edit` and add the following to the file:
```
hpi_openid:
    client_id: my_client_id
    client_secret: my_client_secret
    redirect_uri: my_redirect_uri
```
For development, redirect uri should be `http://localhost:3000/users/auth/hpi/callback`

skydivin4ng3l:
> For setting this up:
> Run `rails credentials:edit` and add the following to the file:
> 
> ```
> hpi_openid:
>     client_id: my_client_id
>     client_secret: my_client_secret
>     redirect_uri: my_redirect_uri
> ```
> For development, redirect uri should be `http://localhost:3000/users/auth/hpi/callback`
> 
> If it throws an error the first time, it just generated your master key. Delete `config/credentials.yml.enc` and rerun `rails credentials:edit`

IF this will be the final Thing, can we have a wiki entry for it. with an explanation in which scenario it has to be done and by whom....for example one time only..by each developer.. is already added to git ignore or other similar things. Also the generation of the client etc could be interesting for this entry.

Issue #238, opened 2019-01-11
Import and Update Puppet scripts


arne-z:
Prerequisite #237 

[//]: # (Lines starting with [//]: are considered comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**
As a User
I want the VMs that were created outside of HART to be accessible to me
in order to use the tool with VMs created in the past.

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

**Acceptance Criteria:**

- [x] All the puppet scripts in the puppet repo are read to find out which users should have access to which VMs.
- [x] The git repo containing puppet scripts is either periodically checked for changes, which are then imported or a live version is used, whichever is better from a development/architecture perspective.
   - Research the correct method of staying up to date with the git version and handling concurrent pushes. 
- [x] The current user access rights in the system should match the Puppet scripts in the repository.
- [ ] Changes to user access in the hart system are immediately saved into the repo.
- [x] Git Repo should remain the single point of truth for user access rights at all times.

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

mewitte:
Tasks:

[ ] Pull current version of repo
[ ] Parse puppet scripts in readable format
[ ] Check existing VMs for updates and edit them
[ ] Check for new VMs

BjoernDaase:
Works and can be closed!

PR #267, opened 2019-01-16
238/vertical/import-update-puppet-scripts

mewitte:
#238


app/api/v_sphere/virtual_machine.rb
```diff
@@ -261,6 +270,66 @@ def status
...
+          all_users = PuppetParserHelper.read_node_file(name)
+          sudo_users = all_users['admins']
+          new_users = User.find_all_by_id(ids)
+          puppetscript = Puppetscript.node_script(name, sudo_users, new_users)
+          git_writer.write_file(name, puppetscript)
+          message = commit_message(git_writer)
+          git_writer.save(message)
+        end
+      rescue Git::GitExecuteError
+        { alert: 'Could not push to git. Please check that your ssh key and environment variables are set.' }
```
Adrian-St: It's a bit pointless to return the error if you're not actually displaying it.
If you don't want to display it, maybe use Rails.logger.info to write the error in the logs.

app/api/v_sphere/virtual_machine.rb
```diff
@@ -261,6 +270,66 @@ def status
...
+          all_users = PuppetParserHelper.read_node_file(name)
+          users = all_users['users']
+          new_sudo_users = User.find_all_by_id(ids)
+          puppetscript = Puppetscript.node_script(name, new_sudo_users, users)
+          git_writer.write_file(name, puppetscript)
+          message = commit_message(git_writer)
+          git_writer.save(message)
+        end
+      rescue Git::GitExecuteError
+          { alert: 'Could not push to git. Please check that your ssh key and environment variables are set.' }
```
Adrian-St: Same here

app/models/user.rb
```diff
@@ -118,6 +128,8 @@ def update_repository
                   end
         git_writer.save(message)
       end
+    rescue Git::GitExecuteError
+      { alert: 'Could not push to git. Please check that your ssh key and environment variables are set.' }
```
Adrian-St: Same here

app/api/v_sphere/virtual_machine.rb
NiklasKoehnecke: this is handeled by #357

PR #282, opened 2019-01-20
Adjust input types for requests and request_templates

MaximilianKoenig:


app/controllers/requests_controller.rb
```diff
@@ -125,6 +125,12 @@ def unsuccessful_action(format, method)
     format.json { render json: @request.errors, status: :unprocessable_entity }
   end
 
+  def prepare_params
+    params[:request][:name] = replace_whitespaces(params[:request][:name])
+    params[:request][:ram_mb] = gb_to_mb(params[:request][:ram_mb].to_i)
```
MrModder: I would add a comment here, to explain why this conversion is necessary.
If you don't know that the view is working in GB and the model is not, this might be confusing

app/views/requests/_form.html.erb
```diff
@@ -35,32 +35,32 @@
...
   <div class="field form-group">
-    <%= form.label :ram_mb, "RAM in MB" %>
+    <%= form.label :ram_mb, "RAM in GB" %>
     <%= form.number_field :ram_mb, min: 0, class: 'form-control', id: 'ram' %>
   </div>
 
-  <div class="field">
-    <%= form.label :storage_mb, "Storage in MB" %>
+  <div class="field form-group">
+    <%= form.label :storage_mb, "Storage in GB" %>
```
MrModder: Are you sure this works?
If the request already has a storage size set, will it not display the value in MB here?

app/helpers/requests_helper.rb
```diff
@@ -4,4 +4,12 @@ module RequestsHelper
   def replace_whitespaces(name)
     name.parameterize(preserve_case: true)
   end
+
+  def mb_to_gb(megabytes)
```
MrModder: Are we talking about GigaBytes and MegaBytes or GibiBytes and MibiBytes? if we are talking about Gibi, we would have to work with 1024 instead of 1000

app/controllers/requests_controller.rb
```diff
@@ -125,6 +125,12 @@ def unsuccessful_action(format, method)
     format.json { render json: @request.errors, status: :unprocessable_entity }
   end
 
+  def prepare_params
```
MrModder: I do like that this is now it's own method
However, you should explicitly check whether params[:request][:name], etc. exists before accessing it.
If they don't at the moment the server will probably error out

app/helpers/requests_helper.rb
BjoernDaase: We are really talking about Megabyte and Gigabyte

Issue #148, opened 2018-12-06
Complete VM to responsable relation (also in vSphere)


BjoernDaase:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

As an employee I want that my requested VMs are saved under the responsables name path (Datacenter >> Datastore >> 'responsables name' >> 'VMName') in vSphere so they are easy to find later. 

**Acceptance Criteria:**

- [x] On the request VM page is a selection filed that contains all existing user (like the selection of users and those with sudo rights)
- [ ] A VM will be created under the first responsables name in vSphere (Datacenter >> Datastore >> 'esponsables name' >> 'VMName') 
- [x] The notification contains that name in the "Responsable" section
- [ ] The name has to be unique, even if two responsables share the same name (e.g. Matthias Müller)

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

BjoernDaase:
Blocked by #163 at the moment.

BjoernDaase:
Looks good to me!

PR #284, opened 2019-01-21
148/scaffold/Complete VM to responsable relation

MrModder:
See #148


app/api/v_sphere/virtual_machine.rb
```diff
@@ -271,10 +271,6 @@ def request
       Request.accepted.find { |each| name == each.name }
     end
 
-    def managed_folder_entry
```
MrModder: Why this change?
It will most likely break inserting VMs into folders!

regnujAx: Because the method exists twice.

MrModder: Oh, that makes sense, GitHub just hid that fact from me 🤦‍♂️

PR #285, opened 2019-01-22
Made redirect uri a config option

jakobmache:
The `redirect_uri` is now a config option, which has to be set with the other credentials. 
I also use the `try` method for accessing values of the `hpi_openid` hash, because if we use standard hash access syntax `[]` and the user has not configured his `credentials.yml.enc` appropriately (`hpi_openid` hash is missing), the server will crash, because it executes `[]` on a `nil` value.


krichly:
We should be able to configure or automatically determine the redirect uri depending on the system/vm, on which the application is deployed. It is necessary to have different redirect uris for the master and dev system, which we could not archive by putting the redirect uri in the encrypted credentials.yml. Possible solutions: find a way to get the base url inside the device initializer or try to configure the hostname in the deployment environments and use it in the device initializer.

jakobmache:
@krichly Would it work by setting the base url in the deploy environment and then exporting it as an environment variable for rails? I did not know how to test this mina locally.

chrisma:
Using the `.try(:[], :client_secret)` construct means that the server will start, regardless whether the necessary openID parameters were found or not. No warning on them not being available and the software therefore not supporting openID logins will be issued.
If openID is considered an essential part of the software, should a misconfiguration that impacts the feature, not be strongly communicated? Having the server not start seems like one way (the naive one) to achieve this.

krichly:
Due to some implementation constraints of the openid identity provider, we have to use two different openid clients for the master and dev systems. That means that we also have to use different client ids and client secrets.

jakobmache:
> Using the `.try(:[], :client_secret)` construct means that the server will start, regardless whether the necessary openID parameters were found or not. No warning on them not being available and the software therefore not supporting openID logins will be issued.
> If openID is considered an essential part of the software, should a misconfiguration that impacts the feature, not be strongly communicated? Having the server not start seems like one way (the naive one) to achieve this.

I would prefer just printing a warning if the credentials are not set.
So in `production` it might be useful if the server does not start when no credentials are configured, but in other environments I don't like this:
- In `development`, this would mean that every developer has to register his own OpenID App, even if he always just develops other parts of the software. Also, every developer would regularly overwrite `credentials.yml.enc`, which is currently in the repo. So every developer has to be aware that he should not commit his personal `credentials.yml.enc`, otherwise it has to be changed again for deployment and every other developer who pulls the updated credentials has to change it again just to start developing. (I think we discussed this yesterday already)
- In `test` we would have to provide credentials for testing or disable the initializer. But I did not find an easy way to disable or change initializers in test mode, the only simple way I saw was to do something like `unless Rails.env.test? #initialization end`. I do not like this approach, because in my opinion for this we have the different environments which are handled by Rails. I also do not see an easy way to provide credentials for testing, because currently there are no environment specific credentials ([but maybe in rails 6](https://github.com/rails/rails/pull/33521)).

So I would prefer just writing a warning to the command line.

> Due to some implementation constraints of the openid identity provider, we have to use two different openid clients for the master and dev systems. That means that we also have to use different client ids and client secrets.

I only see two options here:
- Providing all credentials in one credentials file. But because there are currently no environment specific credentials, we would always need to write something like `Rails.application.credentials.hpi_openid[server][:client_id]` . This also means we have to pass the server the application is running on as an environment variable or with another way to the initializer. But then the application contains in my opinion too much logic which should be done when deploying the app (the app should not know on which server it runs except for necessary information to run like `BASE_URL` to build the `redirect_uri`). The right credentials should be given to the app.
- Overwrite the `credentials.yml.enc` by hand on each server and do not put it into the repo (or have different branches with different versions for each server). This seems to be the cleanest solution for me.

chrisma:
@jakobmache First of all, thanks for your work and commitment in this issue. You make some valuable points. Another important issue is the handling of credentials on the Travis CI instance. Without usage of the `try` construct, the build fails at the `RAILS_ENV=test bundle exec rails db:migrate` stage, as Travis doesn't have access to the unencrypted credentials (https://travis-ci.com/hpi-swt2/vm-portal/builds/98069779). In the light of a quick resolution of this issue, maybe your second solution is easiest. It would then also be useful to register another openID client with the `redirect_uri` set to `localhost:3000/...`, so that this client can be used in development.

krichly:
@jakobmache If we decide for the proposed solution by @chrisma, it is necessary to add the credentials.yml to the gitignore and add the server specific credential.yml files to the shared folder both servers.

chrisma:
@krichly Yup, as it stands, this PR requires there to be separate `credentials.yml.enc` files on the systems deploying the `dev` and `master` branches, which include different OpenID settings for two different OpenID clients, that have the appropriate redirect URIs set (i.e. the `BASE_URL`).

chrisma:
Ein Problem ist noch, dass die `credentials.yml.enc`, die auf den deployten dev und master Servern manuell verwaltet wird, in den `shared` Ordner muss, damit sie beim deployen nicht überschrieben wird.
D.h. sie muss im deploy Schritt aus dem `shared` Ordner dann wieder nach `config/credentials.yml.enc` gesymlinkt werden
Allerdings ist eine `credentials.yml.enc` bereits im git, sodass diese beim clone Schritt des deploy Skripts angelegt wird. Und eine bereits vorhandene Datei kann nicht gesymlinkt werden

Issue #242, opened 2019-01-11
Stop archivings


BjoernDaase:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

<del>As a user whose VM is about to be archived I want to have the possibility to raise an objection in order to stop the archiving process if I still need the VM to be active.</del>  
As a user whose VM is about to be archived I do not want that my VM is archived directly and I do not have the possibility to raise an objection in order to keep my necessary VMs running.
This is also includes an update of the VM details page, to see the state of the VM.

**Acceptance Criteria:**

- [x] There is also a "Confirming Archiving" button on the VM details page
- [x] The shown state of a VM on the VM details page (which is currently only "online" and "offline") also shows the VM archiving process states
- [x] **Show the "Confirm Archiving" button of the admin only after three days**
- [x] If a VM is in the "Confirm Archiving" state on the right site of the status-information there is a button which says "Stop Archiving" which can be clicked by every associated user which stops the archiving and moves the VM to the active ones again 
- [x] The message which is sent to the associated users when a VM is requested to be archived also includes a hint that he can raise an objection if he still needs that VM
- [ ] <del>If a VM is in the "Confirm Archiving" state on the right site of the status-information there is a button which says "Raise Objection". That can be clicked only once per stated user. Furthermore, he has to fill out a field that contains its reason for the objection </del>
- [ ] <del>If a user raised an objection there is a text that says "You raised an objection for this archiving!" instead of the button</del>
- [ ] <del>Furthermore, it also shows the user and there objection reasons which already raised an objection for this archiving to the right</del> 
- [ ] <del>If there are raised objections the admin can not confirm the archiving from the VM overview page. Instead it shows the following text "There are objections to this archiving. You can force the archiving anyway.". Furthermore, the button now says "Force archiving" and shows a pup-up warning.</del>

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

BjoernDaase:
The scroed out lines were removed after the last customer meeting so the workload became less.

BraunTom:
Should the admin also be able to see the Stop Archiving button???

BjoernDaase:
Yes, that would be good.

BjoernDaase:
Looks good to me!

PR #298, opened 2019-01-24
242/scaffold/stop archiving

BraunTom:
#242


app/api/v_sphere/virtual_machine.rb
```diff
@@ -196,7 +241,7 @@ def users
     def root_users
       request = Request.accepted.find { |each| name == each.name }
       if request
-        Request.first.users_assigned_to_requests.select(&:sudo).map(&:user)
+        request.users_assigned_to_requests.select(&:sudo).map(&:user)
```
JonasZim: there is a method in the request model to get the sudo_users

app/models/archivation_request.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ArchivationRequest < ApplicationRecord
+  def can_be_executed?
+    Time.now >= created_at + (60 * 60 * 24 * 3)
```
JonasZim: Please explain these numbers

Issue #308, opened 2019-01-25
Shutdown Guest OS is broken


schmidtflo:
**Summary**

[//]: # (Please provide a details description about the issue you are facing.)

Some of the VM power actions fail. See Screenshot.

![image](https://user-images.githubusercontent.com/4164150/51740556-a5c65d00-2094-11e9-9bdb-348f66ff82ec.png)

Please also include a error handling for power options, so that you will never see a ruby error page.

BjoernDaase:
With #333 being fixed this one can be closed as well.

PR #337, opened 2019-01-29
Fix and handle error with too many resources bug

JoshCoder42:
#308


MrModder:
Such errors should actually already be caught by the requests_controller in the safe_create_vm method
If VM creation throws an error, the user will be informed via an alert stating the error message that was provided by vSphere

MrModder:
Oh, okay, I just noticed that this is with a completely different action
my bad, I thought this was about creating a VM, instead of powering it on. VM creation should already be a safe operation, you can check that code, in the requests_controller maybe it can help you solve your problem

app/controllers/requests_controller.rb
```diff
@@ -199,4 +205,51 @@ def split_requests(requests)
...
+
+    def get_ram_gb(host)
+      host[:summary].hardware.memorySize.to_i / 1024**2
+    end
+
+    def get_storage_gb(host)
+      host[:summary].host.datastore.sum { |datastore| datastore.summary.freeSpace }.to_i / 1024**2
+    end
+
+    hosts = VmApi.instance.all_hosts
```
fyndalf: Using the old `VmApi` here seems to cause timeouts on travis and make the tests slow (and fail without the VPN). Take a look at `app/api/v_sphere` and the classes, as well as the `hosts_controller` or the `vm_controller` and their respective tests to see how the new API is used and can be mocked ;)

fyndalf: `VSphere::Host.all` might do the trick in this case :wink: Check out the `hosts_controller` test and the other host tests to see how it's mocked there :D

app/controllers/requests_controller.rb
```diff
@@ -199,4 +205,51 @@ def split_requests(requests)
     @pending_requests = requests.select(&:pending?)
     @resolved_requests = requests.reject(&:pending?)
   end
+
+  def enough_resources
+    def get_num_cpu(host)
```
fyndalf: Maybe move this method (and the two below) to `api/v_sphere/host` so it can be called more conveniently? There they could also be tested properly :)

spec/controllers/vms_controller_spec.rb
```diff
@@ -226,6 +226,33 @@ def set_old_path
...
+        allow(vm2).to receive(:summary).and_return(summary_double)
+        allow(vm2).to receive(:change_power_state)        
+        
+        post :change_power_state, params: { id: vm2.name }
+      end
+
+      it 'raises an error' do 
+        expect {
+          post :change_power_state, params: { id: vm2.name }
+        }.to raise_error(RbVmomi::Fault)
```
fyndalf: Where would this exception be thrown when the API is mocked? Maybe you can try to figure out a way to get the api request to fail only for this test :)

app/controllers/requests_controller.rb
fyndalf: @JoshCoder42 I took the liberty to mock it, so I'm resolving this conversation :v:

Issue #240, opened 2019-01-11
Add notifications to various events


arne-z:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

As a user of HART I want to receive notifications for various events in order to be notified when my action is required, or an event happens that I care about.  
 

**Acceptance Criteria:**

The following is a list of events, and who they should notify, you can find a PDF with BPMN-diagrams attached at the bottom if you want to know more about the processes behind those notifications.
- [x] New request created -> Admins

- [x] Request accepted or rejected -> Requesting user
- [x] Request accepted -> All users that will have rights on this specific VM

- [x] Archiving of VM requested -> All users with access to the VM and the admins (see page 6 of PDF) (The first text line has to be changed to
"The VM will soon be archived and for that it will then be shut down.
If you still need this VM you can stop the archiving of this VM within three days."
- [x] Archiving of VM requested -> Admins
- [x] VM was archived -> All users that will have rights on this specific VM

- [x] Reviving of VM requested -> Admins after X days
- [x] VM was revived from archive -> All users that will have rights on this specific VM

- [x] The role of a user is changed -> The affected user ("Your role has changed from X to Y on dd.mm.yy" + Link to personal profile page)

- [x] Create a new project -> The project responsables ("The project <ProjectName> with you as the responsable has been created" + Link to the projects page)

Some of the above may already have notifications, please communicate with the other teams to check that all of the above are using notifications and doing so in a consistent fashion.

[Processes.pdf](https://github.com/hpi-swt2/vm-portal/files/2749257/Processes.pdf)

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

BjoernDaase:
Looks good to me and worked in demo perfectly today!

PR #343, opened 2019-02-01
send notifications after various events

regnujAx:
about issue #240


app/controllers/projects_controller.rb
```diff
@@ -29,6 +29,14 @@ def create
       @selected_user_ids = project_params[:responsible_user_ids]
       render :new
     end
+    # nil-check is necessary because 3 tests fails without check
```
MaximilianKoenig: This comment is not helpful. Maybe check whether the tests are correct then.

app/controllers/projects_controller.rb
```diff
@@ -29,6 +29,14 @@ def create
...
+    return if @project.id.nil?
+
+    @project.responsible_users.each do |each|
+      each.notify('Project created',
+                  'The project with you as the responsable has been created: ' +
```
MaximilianKoenig: Typo: responsible instead of responsable

app/controllers/users_controller.rb
```diff
@@ -25,7 +25,12 @@ def update
 
   def update_role
     @user = User.find(params[:id])
+    role = @user.role
```
MaximilianKoenig: Probably rename variable 'role' to 'former_role' or something.

PR #359, opened 2019-02-04
fix notifications

regnujAx:
extend issue #240


app/controllers/requests_controller.rb
```diff
@@ -145,7 +145,12 @@ def notify_request_update
     return if @request.pending?
 
     if @request.accepted?
-      notify_users('Request has been accepted', @request.description_text(host_url))
+      ([@request.user] + User.admin).each do |each|
```
MaxSchneider1337: what do you want to achieve using ([@request.user] + User.admin)?

regnujAx: I only adopt the content of the notify_users method but without the @request.users as the recipients because they receive another notification. Do you think there is something wrong?

Issue #305, opened 2019-01-25
make regular login go away


isabel-amon:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

Now that HPI-Login works, as a user I don't need the regular login anymore. I want the fields to go away so that I am not confused. Also, the buttons "sign up", "forgot your password?" and "sign in with hpi" should not be there as they confuse me.

**Acceptance Criteria:**

- [ ] only the oAuth button is an option to sign in

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

PR #360, opened 2019-02-04
305/login/only hpi login

janwestphal306:
#305


janwestphal306:
Development Mode:
image


Production Mode:
image

We can think about making the HPI-Login Button prettier but it should work for now :D

lassesteffen:
Looks good

PR #368, opened 2019-02-05
Dashboard Cleanup

### LINKED ISSUE #323: Dashboard cleanup
[//]: # (Lines starting with [//]: are considered comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

As a User
I want to have only important information and my own VMs on my dashboard
in order to have a better experience with the product.

**Acceptance Criteria:**

- [ ] Dashboard shows the users own **active** VMs and **to be archived** VMs 
- [ ] To be archived VMs have a red note with "To be archived"
- [ ] No other VMs are shown on the dashboard, neither other users VMs nor Archived VMs
- [ ] Hosts and/or Servers are not included in the dashboard
- [ ] Notifications are shown on the right of the VMs just like they are now
- [ ] Non-functional buttons (like the delete button) should be removed.
- [ ] Instead of the details button linking to the VM's page the name should link to the VMs page (like it does on the VM list)
- [ ] The VM's status should be visible (again, you can do this just like the VMs page)

If possible it would be great if the Dashboard looks a bit cleaner and more like the VM page, but this is not a functional requirement.

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

westphal-jan:
Fixes #323


app/views/dashboard/index.html.erb
```diff
@@ -14,64 +14,96 @@
...
-            <tr>
-              <td><%= vm.name %></td>
-              <td><%= vm.boot_time %></td>
-              <td><%= button_to "Details",
-                                {controller: :vms, action: 'show', id: vm.name},
-                                method: :get, class: "btn btn-success" %></td>
-              <td><%= button_to "Delete",
-                                {controller: :vms, action: 'destroy', id: vm.name},
-                                method: :delete, class: "btn btn-danger" %></td>
+          <% @vms&.each do |vm| %>
```
mlichtblau: Why is there an `&`?

mlichtblau (commented):
I found an `&` character, where I don't know why it's there. Otherwise it looks good to me!
Is the `&` a mistake?

app/views/dashboard/index.html.erb
janwestphal306: `each` will only be called if vms is not null.

mlichtblau: Ok, thanks :)

mlichtblau (approved):

PR #369, opened 2019-02-05
Pull git repo instead of always deleting it

mewitte:
Also fixes error when already existing repo is trying to be cloned

### DIFF
--- a/lib/git_helper.rb
+++ b/lib/git_helper.rb
@@ -10,11 +10,11 @@ module GitHelper
   class GitWriter
     def initialize(path)
       @path = path
-      unless File.exist?(File.join(path, '.git'))
-        setup_git
-      else
+      if File.exist?(File.join(path, '.git'))
         open_git
         @git.pull
+      else
+        setup_git
       end
     end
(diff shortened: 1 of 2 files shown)


lib/git_helper.rb
```diff
@@ -38,15 +39,19 @@ def save(message)
...
-      git.config('user.name', ENV['GITHUB_USER_NAME'])
-      git.config('user.email', ENV['GITHUB_USER_EMAIL'])
-      git
+      @git = Git.clone(uri, name, path: File.join(@path, '..'))
+      @git.config('user.name', ENV['GITHUB_USER_NAME'])
+      @git.config('user.email', ENV['GITHUB_USER_EMAIL'])
+    end
+
+    def open_git
+      @git = Git.open(path)
```
hkohlsaat: vielleicht? `@git = Git.open(@path)`

mewitte: Fixed directly on dev with commit 25e66949052a8f8c273932d01b0c6b47c1d88254

Issue #306, opened 2019-01-25
delete old logins and drop test-users


isabel-amon:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

As a user, now that oAuth works, I no longer need the test logins. If they still appear in the "users" list, I am very confused. 

**Acceptance Criteria:**

- [ ] Only real users are shown in the users list
- [ ] Login with the old test logins doesn't work anymore

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

skydivin4ng3l:
Auf dev ist jetzt, keine seed mehr mit Userdaten. Sinnvoll wäre einmal die Datenbank zu dropen auf DEV und wo es sinnvoll ist, durch das seeden sollten auf jedenfall keine Testuser mehr rein kommen.

skydivin4ng3l:
Habe Christoph mal gebeten, bei nächster Gelegenheit die DB zu droppen auf den Servern. @isabel-amon reicht das für die Erfüllung der Story?

isabel-amon:
wenn dann keine user@user.de etc erscheinen, schon.

skydivin4ng3l:
@BjoernDaase can you make a note here as soon as the DB is dropped?

BjoernDaase:
Of course. Please tell when I should drop it.

BjoernDaase:
I will not drop Dev, I don't see a reason for that.
But Master is dropped completely so there are no more test users.

PR #370, opened 2019-02-05
add "Prefab seed" Test users for different roles, usable in OpenID Testmode

skydivin4ng3l:
add "Prefab" Test users for different roles, usable in OpenID Testmode just comment in the user with the role you which to login with when clicking login with OAuth on Local Dev Server.


skydivin4ng3l:
@jakobmache what am i doing wrong? does not work as intended?

skydivin4ng3l:
@BjoernDaase asked that this PR should be completed, but I can't do it by my own. Can no longer drop DB on local maschine for example and I don't know what I am doing wrong, I thought that this is how @jakobmache explained it.

skydivin4ng3l:
supposed to fix #306 or would should we do this?

krichly:
Not needed!

BjoernDaase:
I also think this PR can be closed because we have the Development LogIn in development mode, where you can choose whether you wan't to be an admin, employee or regular user.

chrisma:
The only changes in this PR are currently 5 lines of comments to ` db/seeds.rb`.
The goal of this PR seems addressed by other PRs, i.e. https://github.com/hpi-swt2/vm-portal/pull/439, so closing for now.

PR #371, opened 2019-02-05
Fix git errors

Adrian-St:


Adrian-St:
@mewitte Thanks for checking, should be fixed now

mewitte:
Yes it did

arne-z:
@Adrian-St will this fix the bug where only one user is added to the init.pp? Or is this a different git error?

app/models/request.rb
```diff
@@ -85,13 +85,13 @@ def create_vm_in_cluster(cluster)
...
   def push_to_git
-    GitHelper.open_repository(PuppetParserHelper.puppet_script_path) do |git_writer|
+    GitHelper.open_repository(Puppetscript.puppet_script_path) do |git_writer|
       git_writer.write_file(File.join('Node', "node_#{name}.pp"), generate_puppet_node_script)
       git_writer.write_file(File.join('Name', "#{name}.pp"), generate_puppet_name_script)
       git_writer.save(commit_message(git_writer))
-    rescue Git::GitExecuteError
-      { alert: "Users could not be associated with the VM!\nCould not push to git. Please check that your ssh key and environment variables are set." }
     end
+  rescue Git::GitExecuteError => e
```
mewitte: Shouldn't this be over the end?

PR #372, opened 2019-02-05
Fix logging in with wrong folder structure in PuppetFiles git repo

mewitte:


mewitte:
This is fixed with https://github.com/hpi-swt2/vm-portal/pull/371

PR #377, opened 2019-02-06
Fix: Write puppetscripts with right names on request.accept

mewitte:
This should fix that the users are not shown on the vms' details pages


mewitte:
https://github.com/hpi-swt2/vm-portal/pull/378 with fresh dev pull

PR #382, opened 2019-02-06
Speed up vms page for user with multiple vms

mewitte:


mewitte:
This is pretty dirty and won't be merged

PR #384, opened 2019-02-06
fix unssuported format in puppet files

arne-z:


lib/puppetscript.rb
```diff
@@ -44,7 +44,7 @@ def self.user_script(user)
 
   def self.generic_node_script
     <<~NODE_SCRIPT
-      class node_%s {
+      class node-%s {
```
mewitte: Inside the file should be an underscore, file name should be with hyphen

lib/puppetscript.rb
```diff
@@ -58,8 +58,8 @@ def self.generic_name_script
     <<~NAME_SCRIPT
       node \'%s\'{
 
-          if defined( node_%s) {
-                      class { node_%s: }
+          if defined( node-%s) {
```
mewitte: Same as above

lib/puppetscript.rb
```diff
@@ -58,8 +58,8 @@ def self.generic_name_script
...
+                      class { node-%s: }
```
mewitte: Same as above

lib/puppetscript.rb
```diff
@@ -69,9 +69,9 @@ def self.generate_user_array(users)
     users = users.map { |user| "\"#{user.email.split('@').first}\"" }
     users.join(', ')
   end
-
+  
   def self.node_file_correct?(vm_name, contents)
-    result =    contents.lines[0].chomp.eql?('class node_' + vm_name + ' {')
+    result =    contents.lines[0].chomp.eql?('class node-' + vm_name + ' {')
```
mewitte: Same as above. If this method returns false, then the Unsupported Format is raised

arne-z:
This was in error. files should have _ in the file and - in the file name.

PR #386, opened 2019-02-06
Only pull git for writing and if not pulled in the last minute

mewitte:


MrModder:
If you want to speed up the processes involving git, I would also recommend caching the results of the VSphere::VirtualMachine#users and VSphere::VirtualMachine#sudo_users
The results could just be saved as instance variables, as VirtualMachine objects are always instanciated on demand an won't exist for long.
Don't forget to update the cached variables in users= and sudo_users= respectively

mewitte:
I don't think caching is needed for the main problem right now. The main problem being that git pull is called once for every VM when loading dashboard index or vms index. From what I've seen that reduces the response time from about 1.3s to 50-60ms. If it is still too slow due to IO then caching might be good to implement.

MrModder:
Why is this the case anyway?
The VMs index page does not contain any information that needs to be pulled from git, except for selecting only the vms the user has access to.
So it should just do one pull. Why does it do one for every vm?

mewitte:
At some point when loading vms index and dashboard index,, _vm_status.html.erb is loaded. This calls vm.sudo_users.

MrModder:
Ah, okay, that makes sense
Thank you

PR #390, opened 2019-02-06
Vertical/minifixes

luisebenkert:
- Restrict access to servers for users
- Restrict access to operating systems for users & employees
- Restrict access to request templates for users & employees

- Remove operating system button on requests page for users & employees
- Remove request templates button on requests page for users & employees

### DIFF
--- a/app/controllers/servers_controller.rb
+++ b/app/controllers/servers_controller.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
 class ServersController < ApplicationController
-  
-  before_action :authenticate_employee, only: %i[show]
+  before_action :authenticate_employee
   before_action :authenticate_admin, only: %i[new create edit update destroy]
   # GET /servers
   # GET /servers.json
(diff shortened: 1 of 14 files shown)


app/controllers/servers_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 class ServersController < ApplicationController
   before_action :set_server, only: %i[show edit update destroy]
-  before_action :authenticate_employee, only: %i[show]
+  before_action :authenticate_employee
   before_action :authenticate_admin, only: %i[new create edit update destroy]
   # GET /servers
   # GET /servers.json
```
BjoernDaase: I think that was correct before.
An employee may see all servers, but may not create a new one or edit existing ones. Only admins are allowed to do so.

BjoernDaase:
The remaining non-test code looks perfect! I can not tell whether the tests are correct.

app/controllers/servers_controller.rb
arne-z: What Bjoern says is correct.

Issue #332, opened 2019-01-28
Move API information to environment


arne-z:
[//]: # (Please use this template for arbitrary tasks which aren't bugs or user stories. This issue type can be used for internal and development f)
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**Description**

The API information that is currently hardcoded into both app/api/vmapi and app/api/vsphere/connection should instead be imported from an environment. 

When you merge this change, please make sure to notify everyone to update their environment file.

[//]: # (Feel free to add new sections if required. Don't forget to add a priority as a label.)

MrModder:
I would actually recommend removing the old API entirely
As far as I know it is no longer used at all and therefore only dead code.

BjoernDaase:
Looks good to me!

PR #392, opened 2019-02-07
Move environment variables into ENV

MrModder:
includes tests and warning to admins/users

#332


MrModder:
I would like to have at least 2 reviewers for this PR, as it may potentially break a lot of things that cannot be tested easily

PR #396, opened 2019-02-11
Capitalize the first character and each that follows a dot or dash

BjoernDaase:


app/models/user.rb
```diff
@@ -76,8 +76,8 @@ def human_readable_identifier
     email.split(/@/).first
   end
 
-  def human_readable_identifer_capitalized
-    self.human_readable_identifier.first.capitalize.gsub(/[-.][a-z]/, &:upcase) #Capitalize the first letter and upcase every letter that follows a . or -
+  def human_readable_identifier_capitalized
```
MrModder: Is there any reason to use an extra method human_readable_identifier_capitalized and not just change the human_readable_identifier method?

BjoernDaase: As we use the human_readable_identifier for vSphere I wasn't sure whether we have restrictions there and wasn't sure whether we can change it.

MrModder: I am pretty sure this can be changed to use a single human_readable_identifier that is simply capitalized
This would be a lot easier to maintain
But let me just check the vSphere docs for a moment

MrModder: The vSphere API docs say the following about Folder names:
```
The name to be given the new folder. An entity name must be a non-empty string of less than 80 characters. The slash (/), backslash (\) and percent (%) will be escaped using the URL syntax. For example, %2F. 
````
So using capitalized characters should not be a problem at all
We might however want to simply cut off the human_readable_identifier at 80 chars, to prevent this from being an esoteric issue later on
The other special characters should not be in the first part of an email anyway so we should not need to worry about it

File: app/models/user.rb [hunk 2]
```diff
@@ -96,7 +96,7 @@ def self.from_omniauth(auth)
 
   def self.from_mail_identifier(mail_id)
     all.each do |user|
-      return user if user.email.split('@').first.casecmp(mail_id).zero?
+      return user if user.human_readable_identifier.casecmp(mail_id).zero?
```
MrModder: definitely keep the change to use the human_readable_identifier here and in the other places though, it is a lot cleaner than before 👍

app/models/user.rb
BjoernDaase: Alright.
We don't have to care about the 80 character limit because we know that the OpenID identifier returns a string with a maximum length of 19 characters (ask @krichly for details). It also only contains the dot and dash as special characters, so we won't have a problem there as well.

PR #397, opened 2019-02-12
Set an explicit version of the Sqlite3 Gem 

BjoernDaase:
Because newer ones (1.4.0, released on 02/05/2019) produce errors


BjoernDaase:
We also updated some dependencies. If this causes error, we have to revert this commit.

PR #402, opened 2019-02-15
399/make settings configurable

### LINKED ISSUE #287: Set the archiving timeout

**User Story**

[//]: # (Be descriptive about your user story. Who requests a feature and what's his intent?)

As an admin I want to have the possibility to set the global period of time after that I get informed about archiving requests in order to set it that it fits my behavior.

**Acceptance Criteria:**

- [ ] It exists an extra admin area
- [ ] That can only be accessed by an admin
- [ ] It can be accessed via a gearwheel left to the notification bell
- [ ] The gear is also only visible for the admin
- [ ] On that the admin can set the "archiving request waiting time"
- [ ] The "Confirm archiving" button and the notification that a VM should be a archived will be sent after that "archiving request waiting time"
- [ ] The "Manage OS" and "Manage Request Templates" buttons are also moved into the admin area

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

### LINKED ISSUE #399: make settings configurable via settings page
**User Story**
As an administrator, I want a new menu item (left to the notification icon). The menu item should be the font awesome cog. This menu item should have a submenu "settings", which links to the settings page.    
The settings page should have the design of the VM request page and should be able to configure the settings for git, vsphere, archiving time, and mailer. 

**Acceptance Criteria:**

- [x] create a model to store settings in the database
- [x] create a menu item with cog icon
- [x] submenu item "settings" links to settings page
- [x] make settings for git, vsphere... configurable
- [x] remove all corresponding environment variables

### LINKED ISSUE #403: move menu items
**Description**
Move "Users" menu item to cog submenu. 
Move "Template" button to cog submenu.
Move "Operating systems" button to cog submenu.   

**Related Issues**
- #402

BjoernDaase:
Fixes #399, also fixes #403 and also closes #287


app/api/v_sphere/connection.rb
```diff
@@ -37,7 +37,7 @@ def clusters_folder
     end
 
     def configured?
-      !@server_ip&.empty? && !@server_user&.empty? && !@server_password&.empty?
+      !@server_ip.nil? && !@server_user.nil? && !@server_password.nil?
```
MrModder: &.empty? was actually intended here, because that would catch both a nil-Configuration and an empty one.
Whats your reason for this change?

BjoernDaase: I think that's the point @BraunTom sees...
image

BjoernDaase: When the value is nil, it evaluates to true.

app/models/archivation_request.rb
```diff
@@ -5,6 +5,10 @@ def self.timeout
     AppSetting.instance.vm_archivation_timeout.days
   end
 
+  def self.timeout_readable
```
MrModder: Good that this is its own method in the ArchivationRequest class 👍

app/api/v_sphere/virtual_machine.rb
```diff
@@ -294,7 +294,7 @@ def users
       rescue Git::GitExecuteError => e
         Rails.logger.error(e)
       end
-      users
+      users || []
```
MrModder: Maybe add a comment here explaining why this change is necessary, because users is already set to [].

app/api/v_sphere/connection.rb
MrModder: Ah, right...
Ok, I need to change that, but we should still check that the strings are not empty, as an empty configuration cannot be valid anyway

app/api/v_sphere/virtual_machine.rb
BraunTom: Moved || [] to the line where nil is assigned to users. Think it is clear enough without a comment now

app/api/v_sphere/connection.rb
BraunTom: Changed it so it checks for nil and empty string

app/api/v_sphere/virtual_machine.rb
```diff
@@ -289,12 +289,12 @@ def users
       begin
         GitHelper.open_repository Puppetscript.puppet_script_path do
           remote_users = Puppetscript.read_node_file(name)
-          users = remote_users['users']
+          users = remote_users['users'] || []
```
MrModder: Aaaaah, thank you, much clearer now 👍

BjoernDaase:
Ok, so we can merge now?

PR #414, opened 2019-02-22
400/configurable vSphere rootfolder

### LINKED ISSUE #400: configurable vsphere root folder
**User Story**

As an administrator, I want to set a vsphere root folder on the settings page. The hart system should create the folder in vsphere center and only manage the vms inside this folder.  

**Acceptance Criteria:**

- [ ] Root folder can be specified on settings page
- [ ] In the vsphere space on the settings page should be the possibility to set a hart root folder.
- [ ] The corresponding folder is created in the vsphere center 
- [ ] Only the vms inside this hart root folder are managed by hart. All other vms are ignored.

MrModder:
Fixes #400

### DIFF
--- a/app/models/app_setting.rb
+++ b/app/models/app_setting.rb
@@ -6,6 +6,9 @@ class AppSetting < ApplicationRecord
   validates :github_user_name, :git_repository_name, :git_repository_url, presence: true
   validates :email_notification_smtp_port, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 65_535 }, allow_nil: true
   validates :vm_archivation_timeout, numericality: { greater_than_or_equal_to: 0 }
+  validates :vsphere_root_folder,
+            length: { less_than: 80 },
+            format: { without: %r{/%\/}, message: 'The vSphere root folder may not contain "/", "\" or "%"' }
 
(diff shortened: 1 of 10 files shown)


app/models/app_setting.rb
```diff
@@ -6,6 +6,9 @@ class AppSetting < ApplicationRecord
   validates :github_user_name, :git_repository_name, :git_repository_url, presence: true
   validates :email_notification_smtp_port, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 65_535 }, allow_nil: true
   validates :vm_archivation_timeout, numericality: { greater_than_or_equal_to: 0 }
+  validates :vsphere_root_folder,
+            length: { maximum: 79 },
```
krichly: Is there a vSphere constraint that limits the folder name to 79 characters?

BjoernDaase: Exactly.

krichly: Can we define a constant VSPHERE_FOLDER_NAME_CHARACTER_LIMIT, which stores the value to avoid magic numbers?

BjoernDaase: Done!

PR #415, opened 2019-02-22
404/Make descriptions searchable

### LINKED ISSUE #404: make description and project of vm searchable
**Description**
It should be possible to filter vms based on the description text and project in the search bar on the vms overview page.

BjoernDaase:
Fixes #404, projects were already searchable

### DIFF
--- a/app/views/vms/index.html.erb
+++ b/app/views/vms/index.html.erb
@@ -2,11 +2,11 @@
   <h1 class="float-left">Virtual Machines</h1>
   <% if current_user.employee_or_admin? %>
     <%= button_to "New Request", new_request_path,
-      method: :get,
-      class: "btn btn-primary float-right mt-3" %>
+                  method: :get,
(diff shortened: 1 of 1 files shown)


app/views/vms/index.html.erb
```diff
@@ -58,6 +59,9 @@
       <td class="table-active align-middle">
         <%= render 'vm_status', vm: vm %>
       </td>
+      <td class="d-none"> <!-- Also not intended to be shown! -->
+        <%= vm.ensure_config.description %>
```
MrModder: Dont use ensure_config here
This is not the place to create config objects, that would be the controller.
use config instead and do a nil check with
@vm.config&.description
Otherwise we would create unneccessary config objects

BjoernDaase: Thanks, changed it.

rteusner:
lgtm 👍

PR #417, opened 2019-02-24
411/Add dynamic PuppetScript Settings

### LINKED ISSUE #411: add puppet settings to settings page
**Description**

There should be a section Puppet, where I can specify the following three fields:

init.pp path
classes folder path
nodes folder path 

**Related Issues**

- #399

BjoernDaase:
Fixes #411


app/api/v_sphere/virtual_machine.rb
```diff
@@ -58,9 +58,10 @@ def self.find_by_name(name)
     end
 
     def self.prepare_vm_names
-      return [] unless File.exist?(File.join(Puppetscript.puppet_script_path, 'Node'))
+      node_file = Puppetscript.nodes_path
```
MrModder: Should this not be node_dir/node_path instead of node_file?

db/migrate/20190224103931_add_git_puppet_settings.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddGitPuppetSettings < ActiveRecord::Migration[5.2]
+  def change
+    add_column :app_settings, :puppet_init_path, :string, default: ''
```
MrModder: The defaults here should probably match the defaults when creating the app_settings instance

lib/puppetscript.rb
```diff
@@ -98,4 +98,28 @@ def self.read_node_file(vm_name, repository_path = puppet_script_path)
   def self.puppet_script_path
     GitHelper.repository_path
   end
+
+  def self.init_file_name
```
MrModder: Much better, no more magic values!

app/api/v_sphere/virtual_machine.rb
BjoernDaase: Already fixed.

db/migrate/20190224103931_add_git_puppet_settings.rb
BjoernDaase: You means like /Nodes, /Names and /Init?

BjoernDaase: Done!

app/api/v_sphere/virtual_machine.rb
```diff
@@ -58,9 +58,10 @@ def self.find_by_name(name)
     end
 
     def self.prepare_vm_names
-      return [] unless File.exist?(File.join(Puppetscript.puppet_script_path, 'Node'))
+      nodes_path = Puppetscript.nodes_path
```
krichly: How is the behavior defined if the folder not exist? In this case, the system should create the corresponding folder.     

Also the user should specify the path to the folder, in which the node/class files are stored (not to the folder where the node/classes folder is).

MrModder: What do you mean by saying
```
Also the user should specify the path to the folder, in which the node/class files are stored (not to the folder where the node/classes folder is)
```
Do you mean, when the user enters "/" as the path for the node folder, that it is automatically converted to "/Node"?
Same for the classes folder.
You are also now talking about the classes folder. This is currently the "/Name" folder. Should this be renamed to "/Class" or something similar?

krichly: My error, everything is fine with the paths handling.

PR #436, opened 2019-02-27
admins can configure ip and dns of vms

### LINKED ISSUE #431: Edit Ip and dns
**Description**

For an administrator, it should be possible to edit the dns and ip for existing vms.

BraunTom:
should close #431


app/views/vms/edit.html.erb
```diff
@@ -22,6 +22,18 @@
       <%= form.text_area :description, cols: "40", rows: "4", class: 'form-control', value: @description %>
   </div>
 
+  <% if current_user.admin? %>
```
BjoernDaase: Is there a reason why you do not change the FormBuilder to the new HartFormBuilder an use the the labeled_text_field?

app/controllers/vms_controller.rb
```diff
@@ -46,6 +46,8 @@ def edit
     @sudo_user_ids = @vm.sudo_users.map(&:id)
     @non_sudo_user_ids = @vm.users.map(&:id)
     @description = @vm.ensure_config.description
+    @ip = @vm.ensure_config.ip
```
MrModder: please use a single @configuration @vm_config or similar instance variable, instead of creating more local variables

app/controllers/vms_controller.rb
```diff
@@ -55,6 +57,8 @@ def update
     @vm.sudo_users = info_params[:sudo_user_ids]
     @vm.users = info_params[:non_sudo_user_ids]
     @vm.ensure_config.description = info_params[:description]
+    @ip = @vm.ensure_config.ip = info_params[:ip]
+    @dns = @vm.ensure_config.dns = info_params[:dns]
```
MrModder: Here we can conform to rails convention and use @vm.ensure_config.update(config_params)

app/controllers/vms_controller.rb
```diff
@@ -189,7 +193,7 @@ def authorize_vm_access
   end
 
   def info_params
-    params.require(:vm_info).permit(:description, sudo_user_ids: [], non_sudo_user_ids: [])
+    params.require(:vm_info).permit(:description, :ip, :dns, sudo_user_ids: [], non_sudo_user_ids: [])
```
MrModder: Maybe create a second function for the VirtualMachineConfig params. Then we could use the rails update method in update.

spec/views/vms/edit.html.erb_spec.rb
```diff
@@ -7,7 +7,10 @@
     v_sphere_vm_mock 'VM', vm_ware_tools: 'toolsInstalled'
   end
 
+  let(:current_user) { FactoryBot.create :user }
```
MrModder: nice solution to call this :current_user

app/controllers/vms_controller.rb
```diff
@@ -56,9 +54,7 @@ def update
     notify_changed_users(@vm.users.map(&:id), info_params[:non_sudo_user_ids].map(&:to_i), false, @vm.name)
     @vm.sudo_users = info_params[:sudo_user_ids]
     @vm.users = info_params[:non_sudo_user_ids]
-    @vm.ensure_config.description = info_params[:description]
-    @ip = @vm.ensure_config.ip = info_params[:ip]
-    @dns = @vm.ensure_config.dns = info_params[:dns]
+    @vm.ensure_config.update config_params
     unless @vm.config.save
```
MrModder: The update method already does a save to the database and returns whether it was successful.
So use the update in the unless statement directly, instead of first updating, then saving.

app/controllers/vms_controller.rb
```diff
@@ -153,10 +153,6 @@ def reset_vm
     redirect_back(fallback_location: root_path)
   end
 
-  # This controller doesn't use strong parameters
-  # https://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html
-  # Because no Active Record model is being wrapped
```
chrisma: Why was this comment removed? Is it no longer relevant?

MrModder: The controller now uses strong parameters
and it also has an ActiveRecordModel now (even though its name does not quite match (vms_controller vs VirtualMachineConfig)

chrisma: Thanks for the info!

chrisma:
Can we still address the comments by @MrModder ?

app/views/vms/edit.html.erb
chrisma: This issue is addressed right? The conversation can be resolved?

chrisma:
There is one more comment by @BjoernDaase that could be addressed, then this PR seems ready for merging :100:

app/views/vms/edit.html.erb
BjoernDaase: Yes.

Issue #463, opened 2019-03-11
Refactor routes.rb


chrisma:
* `root`seems to be routed twice
* Little usage of scopes
* Little usage of Rails resource helpers
* Little structure
* Why are params in routes constrained to `{ id: /.*/ }`? Are params strings and not ids?

chrisma:
the route to reject a request was defined as `patch '/vms/requests/reject', to: 'requests#reject', as: 'reject'`, now it's
```
  scope 'vms' do
    resources :requests, path: 'requests' do
      post :push_to_git, on: :member
      patch :reject, on: :collection
    end
    resources :requests, path: 'requests' do
      patch :reject, on: :collection
    end
end
```

chrisma:
Fixed in https://github.com/hpi-swt2/vm-portal/pull/464

PR #464, opened 2019-03-11
#463/routes refactor

chrisma:
Addresses https://github.com/hpi-swt2/vm-portal/issues/463

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -57,9 +57,7 @@ Rails.application.routes.draw do
 
   resources :hosts, :servers
   resources :users do
-    member do
-      patch :update_role
-    end
+    patch :update_role, on: :member
   end
 
   resources :projects, only: %i[index show new edit create update]
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -40,9 +40,9 @@ Rails.application.routes.draw do
     resources :requests do
       post :push_to_git, on: :member
       patch :reject, on: :collection
-      # '/vms/requests/operating_systems/...'
-      resources :operating_systems, except: :show
     end
+    # '/vms/requests/operating_systems/...'
+    resources :operating_systems, path: 'requests/operating_systems', except: :show
     # '/vms/configs/:id'
     scope 'configs' do
       get ':id' => 'vms#edit_config', constraints: { id: /.*/ }, as: :edit_config
@@ -67,6 +67,7 @@ Rails.application.routes.draw do
     end
   end
 
+  # '/users/...'
   # https://github.com/plataformatec/devise#configuring-routes
   devise_for :users,
     path: 'users',
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,9 +16,11 @@ Rails.application.routes.draw do
   end
 
   resources :notifications, only: %i[index new create destroy] do
-    get :mark_as_read, on: :member
+    member do
+      get :mark_as_read
+      delete :destroy_and_redirect
+    end
     get :has_any, on: :collection, to: 'notifications#any?'
-    delete :destroy_and_redirect, on: :member
   end
 
   get '/dashboard' => 'dashboard#index', as: :dashboard
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,13 +12,12 @@ Rails.application.routes.draw do
 
   # https://guides.rubyonrails.org/routing.html#prefixing-the-named-route-helpers
   scope 'vms' do
-    resources :operating_systems, path: 'requests/operating_systems', except: :show
     resources :request_templates, path: 'request_templates', except: :show
-  end
-
-  resources :requests, path: '/vms/requests' do
-    post :push_to_git, on: :member
-    patch 'reject', on: :collection
+    resources :operating_systems, path: 'requests/operating_systems', except: :show
+    resources :requests, path: 'requests' do
+      post :push_to_git, on: :member
+      patch :reject, on: :collection
+    end
   end
 
   resources :notifications, only: %i[index new create destroy] do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,9 +10,9 @@ Rails.application.routes.draw do
   resources :operating_systems, path: '/vms/requests/operating_systems', except: :show
 
   resources :request_templates, path: '/vms/request_templates', except: :show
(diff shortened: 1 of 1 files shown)


chrisma:
[routes_dev.txt](https://github.com/hpi-swt2/vm-portal/files/2953259/routes_dev.txt)
[routes_refactor.txt](https://github.com/hpi-swt2/vm-portal/files/2953260/routes_refactor.txt)

chrisma:
Fixes #463

config/routes.rb
```diff
@@ -2,60 +2,81 @@
...
+      end
     end
   end
 
-  resources :projects, only: %i[index show new edit create update]
-
-  root 'landing#index'
+  # '/users/...'
+  # https://github.com/plataformatec/devise#configuring-routes
+  devise_for :users, path: 'users',
```
krichly: Why do we need path: 'users'?

config/routes.rb
```diff
@@ -2,60 +2,81 @@
...
+  # https://guides.rubyonrails.org/routing.html#using-root
+  root 'landing#index'
 
-  resources :request_templates, path: '/vms/request_templates', except: :show
-  patch '/vms/requests/reject', to: 'requests#reject', as: 'reject'
-  resources :requests, path: '/vms/requests' do
-    post :push_to_git, on: :member
-  end
+  # '/dashboard'
+  get '/dashboard' => 'dashboard#index', as: :dashboard
```
krichly: Do we need a dashboard controller or is a dashboard method inside the application controller also suitable?

config/routes.rb
```diff
@@ -2,60 +2,81 @@
...
+  # '/projects/...'
+  resources :projects, only: %i[index show new edit create update]
```
krichly: Use except: :destroy instead of only?

config/routes.rb
```diff
@@ -2,60 +2,81 @@
...
+  # '/app_settings/...'
+  resources :app_settings, only: %i[update edit]
+  # '/hosts/...', '/servers/...'
+  resources :hosts, :servers
+  get '/hosts/:id' => 'hosts#show', constraints: { id: /.*/ }
```
krichly: Is this line not be included in the resources command?

config/routes.rb
```diff
@@ -2,60 +2,81 @@
...
-      patch :update_role
+  # https://guides.rubyonrails.org/routing.html#prefixing-the-named-route-helpers
+  # '/vms/...'
+  scope 'vms' do
+    # '/vms/request_templates/...'
+    resources :request_templates, except: :show
+    # '/vms/requests/operating_systems/...'
+    # Order matters here, as routes are matched from top to bottom
+    # If 'resources :requests' is first, it will attempt to match that
+    resources :operating_systems, path: 'requests/operating_systems', except: :show
```
krichly: Why is this nested under vms?

config/routes.rb
chrisma: A dashboard method inside the application controller would totally work. The `app/controllers/dashboard_controller.rb` currently contains, in addition to the `index` methods, some further private methods, e.g. `initialize_vm_categories` and `filter_vm_categories`, which seem duplicated in `app/controllers/vms_controller.rb`. I'd rather not do this refactoring in this PR. I've opened another issue for this.

config/routes.rb
chrisma: I believe the original intent of this line was the `, constraints: { id: /.*/ }`, which allows arbitrary strings as ids. Good questions why this should only be possible for the `show` actions of hosts and not the others. Should URLs for hosts refer to host objects with names? Or can the usual numerical ids be ussed?

config/routes.rb
chrisma: That was the configuration before the refactoring. We can totally change this, maybe not in this refactoring PR though. I'll have to check if any tests / code relies on the OSs being available under that path.

config/routes.rb
krichly: top

config/routes.rb
krichly: I think usual numerical ids are completly fine.

chrisma: I agree that numerical ids would actually be preferable to using names in URLs. However, the current system relies on VM names (e.g. [https://hart-dev.epic-hpi.de/vms/vm/many-responsible](https://hart-dev.epic-hpi.de/vms/vm/many-responsible)), presumably because the vSphere API does not return ids for VMs?

chrisma: I agree that numerical ids would actually be preferable to using names in URLs. However, the current system relies on VM names (e.g. [https://hart-dev.epic-hpi.de/vms/vm/many-responsible](https://hart-dev.epic-hpi.de/vms/vm/many-responsible)) (presumably because the vSphere API does not return ids for VMs?). The use of vm names in URLs in the current system is the explanation for the separate `constraints: { id: /.*/ }` code, allowing this, where usually only numerical ids would be allowed.

config/routes.rb
chrisma: :+1:

config/routes.rb
chrisma: I believe 'users' is the default for Devise anyway, not too bad to have it in the `routes.rb` as a reminder for the employed path and where to change it.

https://www.rubydoc.info/github/plataformatec/devise/ActionDispatch%2FRouting%2FMapper:devise_for

config/routes.rb
chrisma: Related ticket: https://github.com/hpi-swt2/vm-portal/issues/480

config/routes.rb
chrisma: Related ticket: https://github.com/hpi-swt2/vm-portal/issues/480

BjoernDaase:
One thing I had in mind for a longer an I get reminded of it in this MR: Why do we not allow to destroy projects? Is there never a point we might have to many of them and they have to be cleaned up?

config/routes.rb
krichly: That was the reason for my comment, but the argumentation to define the default value as a reminder I do not understand.

chrisma:
> Why do we not allow to destroy projects? Is there never a point we might have to many of them and they have to be cleaned up?

Good question. Ticket: https://github.com/hpi-swt2/vm-portal/issues/488

PR #465, opened 2019-03-11
Debug/452

BjoernDaase:
For testing purpose only.


chrisma:
By default, `<button>` elements are submit buttons. This can be changed by setting `type="button"` or by using a different HTML element and styling it as a button with [Bootstrap CSS classes](https://getbootstrap.com/docs/4.0/components/buttons/#button-tags).

See: https://stackoverflow.com/questions/9824808/disable-form-auto-submit-on-button-click/9825224

chrisma:
See my two commits in this branch for a solution approach.

PR #469, opened 2019-03-12
409/configurable vm setting

BraunTom:


BraunTom:
You probably want to change the values in the seeds.rb. I am not sure which values are sensible

app/views/app_settings/_form.html.erb
```diff
@@ -29,6 +29,22 @@
       </div>
     </div>
 
+    <div class="row border-bottom mb-4 pb-4">
+      <div class="col">
+        <h5>VM Settings</h5>
+      </div>
+      <div class="col-6">
+        <%= form.labeled_text_field 'Minimum cpu core count(GB)', :min_cpu_cores %>
```
BjoernDaase: The GB here and in line 39 are copy & paste artifacts, right?

BraunTom: Yes

db/migrate/20190224103931_add_git_puppet_settings.rb
```diff
@@ -2,8 +2,8 @@
 
 class AddGitPuppetSettings < ActiveRecord::Migration[5.2]
   def change
-    add_column :app_settings, :puppet_init_path, :string, default: '/'
```
chrisma: Why was an existing migration edited instead of creating a new one, which removes the defaults?

The migration will not be applied on a system that has already applied the unchanged `db/migrate/20190224103931_add_git_puppet_settings.rb` migration, e.g. the dev deployment.
Changing migrations requires everyone to drop and recreate the db, working against the idea of migrations.

BraunTom: I just put the default in the seeds.rb and there is no need for changes in the already existing database (I do not want a new value, I just wanted to move where it is initialized) and in the future it should be initialized with the seed if it is created new.

app/Validators/vm_validator.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class VmValidator < SimpleDelegator
```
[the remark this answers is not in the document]
BraunTom: I am not sure what comment you want. 
They share a validator because there was code duplication and I would not want to write:
requests and request_templates share a validator because they both have to check the same vm values
because everytime someone uses this validator he would have to change the comment

app/models/app_setting.rb
```diff
@@ -17,15 +17,7 @@ class AppSetting < ApplicationRecord
...
-    first_or_create!(singleton_guard: 0,
-                     github_user_name: 'MyUserName',
-                     github_user_email: 'example@email.com',
-                     git_repository_name: 'repository',
-                     git_repository_url: 'https://github.com/hpi-swt2/vm-portal.git',
-                     puppet_init_path: '/',
-                     puppet_nodes_path: '/Node',
-                     puppet_classes_path: '/Name',
-                     vm_archivation_timeout: 3) # in days
+    find(1)
```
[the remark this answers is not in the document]
BraunTom: Without seed you cannot log in (the application already wants to use some values of AppSetting at this moment). Therefore I changed it so that you get a message to seed instead of the button to log in. Because of that the admin have to seed the application before the first use and we have a valid state.
A side effect  is that find(1) should always be ok

(first_or_create!() would not be enough without give it almost all the values that are in the seeds.rb)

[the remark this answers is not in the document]
chrisma: :+1:

app/Validators/vm_validator.rb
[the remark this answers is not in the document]
chrisma: Fair enough. Thanks for the reply.

db/migrate/20190224103931_add_git_puppet_settings.rb
chrisma: > there is no need for changes in the already existing database
Good point.

chrisma:
> You probably want to change the values in the seeds.rb. I am not sure which values are sensible

We can address this later. Good to have them in the `seeds.rb` as placeholders for now.

PR #472, opened 2019-03-13
449/created VMs in wrong network

### LINKED ISSUE #449: created VMs in wrong vsphere network
Beim automatischen Erstellen von VMs wird die NIC der Maschine im Netzwerk 'VM Network' eingerichtet. Da dieses Netzwerk nicht existiert, wird es automatisch von vSphere erstellt. Es gibt aber ein Standard Netzwerk, in welchem alle VMs erstellt werden sollen.
Die vSphere API sollte eine Liste zurückliefern und das oberste Element ausgewählt werden. Ansonsten funktioniert das automatische Deployment und der pxeboot nicht.

MrModder:
Closes #449

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -84,27 +84,33 @@ class Request < ApplicationRecord
     clusters = VSphere::Cluster.all
     return nil, 'VM could not be created, as there are no clusters available in vSphere!' if clusters.empty?
 
-    warning = nil
-    begin
-      push_to_git
-    rescue Git::GitExecuteError => e
-      logger.error(e)
-      warning = "Your VM was created, but users could not be associated with the VM! Push to git failed, error: \"#{e.message}\""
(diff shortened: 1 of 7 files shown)


app/models/request.rb
```diff
@@ -84,27 +84,33 @@ def create_vm
...
+      description: description
+    )
     vm.move_into_correct_subfolder
     vm
   end
 
-  # Error handling has been moved into create_vm to provide easier feedback for the user
+  def push_to_git_with_warnings
+    push_to_git
+    nil
```
BjoernDaase: I don't understand this nil here. Can you explain me what it does?

MrModder: if everything works, return nil as warning

MrModder: nil here means, there was no warning

Issue #443, opened 2019-03-06
Refactor ActiveRecord usage in dashboard controller


chrisma:
`app/controllers/dashboard_controller.rb`

```ruby
@notifications = Notification.where(user: current_user).slice(0, number_of_notifications)
```
loads all notifications and then discards. Instead: https://guides.rubyonrails.org/active_record_querying.html#limit-and-offset

chrisma:
`max_shown_vms` and `number_of_notifications` might not be necessary.

Idea: Return all items and hide the majority in the UI...

PR #473, opened 2019-03-13
425/v sphere vm name

### LINKED ISSUE #425: vSphere vm name
**User Story**
As an administrator, I want that the vm names in the vSphere Center include the creation date in order to have this information (e.g., how old a vm is) also in vSphere available.

naming convention: `<yyyy><mm><dd>_vm-<vmname>`, e.g. `20190301_vm-myvm`
`<vmname>` should be lowercase

**Acceptance Criteria:**

- [ ] the system is able to create and access vms with the specified naming convention

BraunTom:
Closes #425


app/views/dashboard/index.html.erb
```diff
@@ -5,9 +5,9 @@
 </div>
 <div class="row">
   <div class="col-8">
-    <%= render 'widget', {vms: @vms, id: 'vms_active', header: 'Active VMs', buttons: nil} %>
-    <%= render 'widget', {vms: @archivable, id: 'vms_to_be_archived', header: 'Archivable VMs', buttons: nil} %>
-    <%= render 'widget', {vms: @revivable, id: 'vms_to_be_revived', header: 'Revivable VMs', buttons: nil} %>
+    <%= render 'widget', {vms: @vms.slice(0...max_shown_vms), id: 'vms_active', header: 'Active VMs', buttons: nil} %>
+    <%= render 'widget', {vms: @archivable.slice(0...max_shown_vms), id: 'vms_to_be_archived', header: 'Archivable VMs', buttons: nil} %>
+    <%= render 'widget', {vms: @revivable.slice(0...max_shown_vms), id: 'vms_to_be_revived', header: 'Revivable VMs', buttons: nil} %>
```
chrisma: `.limit(max_shown_vms)` in `app/controllers/dashboard_controller.rb`?
See https://github.com/hpi-swt2/vm-portal/issues/443 and https://guides.rubyonrails.org/active_record_querying.html#limit-and-offset

app/helpers/dashboard_helper.rb
```diff
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
 
 module DashboardHelper
+  def max_shown_vms
+    10
```
chrisma: We already have the concept of `AppSettings`. Maybe this could be applied here as well?

app/api/v_sphere/folder.rb
```diff
@@ -70,13 +70,17 @@ def move_here(folder_entry)
...
-      vm_config = creation_config(cpu, ram, capacity, name)
+      vm_config = creation_config(cpu, ram, capacity, prefix(name))
       vm = @folder.CreateVM_Task(config: vm_config, pool: cluster.resource_pool).wait_for_completion
       VSphere::VirtualMachine.new vm
     end
 
     private
 
+    def prefix(name)
+      "#{Time.now.strftime('%Y%m%d')}_vm-#{name}"
```
chrisma: This method returns the full name with prefix of a vm, right? `vm_name_with_prefix(name)`, `add_prefix(name)`, `format_vm_name(name)`?
Why is this in `folder.rb`?

app/views/dashboard/index.html.erb
BraunTom: Vms are not a rails model but implemented by us (they wrap the vms from vSphere and are not in the database) so there is no limit method. Furthermore the problem is that we maybe do not want to show the first 10 vms out of all vms (users cannot see all vms).
There is the alternative to slice in the controller, but than there would be a third method with were vms, archivable, revivable are redefined

app/api/v_sphere/folder.rb
BraunTom: It is in the folder.rb because vms are created here
It is just a helper method

app/views/dashboard/index.html.erb
chrisma: Thanks for the reply.
So am I understanding it correctly, that every  API request for VMs in vSphere always returns all VMs and all filtering/slicing/limiting is done in the controllers?

app/api/v_sphere/folder.rb
chrisma: > It is in the folder.rb because vms are created here

Let me rephrase my question: Is there a specific reason why methods for creating VMs (and associated helper methods) are in file called `folder.rb` and not in e.g. `app/api/v_sphere/virtual_machine.rb`? I would have expected VM creation to have to do more with VMs and less with folders.

MrModder: > Let me rephrase my question: Is there a specific reason why methods for creating VMs (and associated helper methods) are in file called folder.rb and not in e.g. app/api/v_sphere/virtual_machine.rb? I would have expected VM creation to have to do more with VMs and less with folders.

The VM creation is in folder.rb because the vSphere's API-call to create Virtual Machines is only available on folders. For the simple reason, that you always need a folder to create the vm in. We simply kept this convention.
And we kept the associated helper methods in the same file to make them easy to find, associate them with the action they are helping and because virtual_machine.rb is already quite large.

chrisma: > vSphere's API-call to create Virtual Machines is only available on folders
Ah, I see, that is the critical information. Thanks.

app/helpers/dashboard_helper.rb
BraunTom: Changed it to use AppSetting

chrisma: :+1:

PR #475, opened 2019-03-14
Layout changes for index pages

chrisma:
* Try to remove as much custom styling as possible
* Return to default Bootstrap styles
* Use Bootstrap grid system
* Employ title attributes and tooltips when possible

PR #479, opened 2019-03-14
Use a better green

BjoernDaase:
Use a better green that is not so bright.


chrisma:
Had the same idea :) https://github.com/hpi-swt2/vm-portal/pull/475/commits/28dbe06bfcca07e85c40fc8a69394bd8a726b003

PR #481, opened 2019-03-16
432/show vSphere path in VMs show

### LINKED ISSUE #432: Display vsphere folder path on vm show page
**Description**

The system should display the vsphere folder path on the vm show page in the section "Hardware / Network".

### LINKED ISSUE #438: Refactor custom ip validations for server model
**Description**

Replace the unreadable regular expression by using a gem: https://stackoverflow.com/questions/3634998/how-do-i-check-whether-a-value-in-a-string-is-an-ip-address

### LINKED ISSUE #477: Management & Caching of AppSettings
Currently, every `AppSettings.instance` call results in a db query... Those are cached, but maybe https://github.com/huacnlee/rails-settings-cached could help.

MrModder:
Closes #432, fixes #438 and also closes #477

### DIFF
--- a/app/models/app_setting.rb
+++ b/app/models/app_setting.rb
@@ -16,7 +16,7 @@ class AppSetting < ApplicationRecord
 
   after_commit :apply_settings, on: :update
 
-  def self.refresh_instance
+  def self.clear_cache
     @instance = nil
   end
 
@@ -28,7 +28,7 @@ class AppSetting < ApplicationRecord
   private
 
   def apply_settings
-    self.class.refresh_instance
+    self.class.clear_cache
     apply_mail_settings
     GitHelper.reset
   end
--- a/app/models/app_setting.rb
+++ b/app/models/app_setting.rb
@@ -21,7 +21,7 @@ class AppSetting < ApplicationRecord
   end
 
   def self.instance
-    @instance = find(1) if @instance.nil
+    @instance = find(1) if @instance.nil?
     @instance
   end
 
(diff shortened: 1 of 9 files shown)


app/views/vms/show.html.erb
```diff
@@ -7,7 +7,13 @@
 end %>
 
 <div class="container mb-3 pl-0 pr-0 clearfix">
-  <h1 class="mb-0"><%= @vm.name unless @vm.nil? %></h1>
+  <ol class="breadcrumb" aria-label="The Virtual Machines full path">
```
BjoernDaase: I like the positioning on the page :+1:

app/models/app_setting.rb
```diff
@@ -16,13 +16,19 @@ class AppSetting < ApplicationRecord
 
   after_commit :apply_settings, on: :update
 
+  def self.refresh_instance
+    @instance = nil
```
chrisma: Would `clear_cache` be a better name for this method?

app/models/server.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+# Part of Ruby stdlib, used for IP regexes
+# https://www.rubydoc.info/stdlib/resolv/Resolv
+require 'resolv'
```
chrisma: Nice use of the Ruby stdlib :+1:

app/models/app_setting.rb
chrisma: https://github.com/hpi-swt2/vm-portal/pull/481/commits/1713c798863ae4602c06abbacf66848ac00a7a7f

PR #484, opened 2019-03-17
480/Cleanup tests

### LINKED ISSUE #460: Deal with pending test cases
Current test suite: `618 examples, 0 failures, 9 pending`

* Investigate pending tests
* Decide whether to keep and include the pending tests or remove them if they are not necessary

BjoernDaase:
Fixes #460, makes boot_time of a VM accessible via the summary and fixes a vSphere root folder validation bug


spec/models/app_setting_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'rails_helper'
 
+# We should add some tests here, which test that the validation is correct
```
chrisma: This should be the last thing for this PR, right?

BjoernDaase: Correct!

BjoernDaase: Done :)

PR #493, opened 2019-03-18
447/generatePuppetScriptsWithTemplating

### LINKED ISSUE #447: Replace dynamic generation of puppet scripts with templating
Puppet scripts are currently being dynamically generated: `https://github.com/hpi-swt2/vm-portal/blob/dev/lib/puppetscript.rb`

This means that code is fairly hard to read and maintain.

Instead, consider using templates for puppet scripts, i.e. `puppet.erb.pp` that is then filled with variables just like RoR views or RoR mailers are.

BraunTom:
Fixes #447


BjoernDaase:
I would suggest moving the `*Script.erb` files not in a `assets` folder, instead I would prefer to have a `puppetscript` folder.  
That would make there intention clearer.

chrisma:
> I would suggest moving the `*Script.erb` files not in a `assets` folder, instead I would prefer to have a `puppetscript` folder.

Agreed. `puppetscript` or  `templates`.

PR #498, opened 2019-03-22
Standardize styling

chrisma:
* Remove as much custom CSS as possible
* Rely on bootstrap classes and layouts
* Remove as much custom JS as possible
* Remove empty coffee and SCSS files
* Restructure Gemfile


chrisma:
> Is it intended, that the user select is that small?
I've added the correct Bootstrap class, so that the field has the same width as the other elements now, good catch. The font size is the default.

PR #501, opened 2019-03-23
Notification rework

chrisma:
* Buttons for notifications: marking as read and deleting
* Show datetime of notification, including human-readable format
* Restrict clickable area of notifications to title
* When title is clicked, mark notification as read
* Display only unread notifications on dashboard

- [ ] Write a test that tests clicking on a `sudo rights granted` notification (i.e. calling `mark_as_read_and_redirect`) and runs into "too many redirects"

image


BjoernDaase:
Clicking on notifications sometimes results in too many redirects for me
image

Sadly, I can not say when it exactly happens...

chrisma:
> Sadly, I can not say when it exactly happens...

Thanks for checking it out. Which type of notification is behaving this way, e.g. 'request accepted', 'role change'?

BjoernDaase:
@chrisma It happend with a "Sudo rights granted" notification.
Replacing it seems to fix the issue :+1:

chrisma:
> @chrisma It happend with a "Sudo rights granted" notification.
> Replacing it seems to fix the issue

While this is a fix, it also diminishes functionality somewhat. I would be interested in a regression test that involves a `sudo rights granted` notification.

BjoernDaase:
> > @chrisma It happend with a "Sudo rights granted" notification.
> > Replacing it seems to fix the issue
> 
> While this is a fix, it also diminishes functionality somewhat. I would be interested in a regression test that involves a `sudo rights granted` notification (and the original code).

We should definitely think about it!

chrisma:
Even with empty links on notifications, I cannot reproduce this error. In the Rails console: `User.first.notify('My test notification', 'This does not have a link')`

chrisma:
As I'm currently not able to reproduce the issues, neither manually nor with the automated tests, I'm merging this, so that development can continue with these changes.

However, please open another issue if errors are found.

PR #502, opened 2019-03-23
488/DeleteProjects

### LINKED ISSUE #488: Deletion of projects
> Why do we not allow to destroy projects? Is there never a point we might have to many of them and they have to be cleaned up?

Most likely it should be possible to delete projects.

BjoernDaase:
Fixes #488


app/controllers/projects_controller.rb
```diff
@@ -49,6 +49,14 @@ def update
     end
   end
 
+  def destroy
+    @project.destroy
+    respond_to do |format|
+      format.html { redirect_back fallback_location: projects_url, notice: 'Project was successfully deleted.' }
+      format.json { head :no_content }
```
chrisma: Only HTML

chrisma: Only HTML

db/schema.rb
```diff
@@ -37,8 +37,8 @@
     t.integer "max_cpu_cores"
     t.integer "max_ram_size"
     t.integer "max_storage_size"
-    t.integer "max_shown_vms"
     t.string "git_branch", default: "master"
+    t.integer "max_shown_vms", default: 10
```
chrisma: No migrations in this PR, changes to schema.rb not necessary

app/views/projects/_list.html.erb
```diff
@@ -0,0 +1,33 @@
+<table class="table table-striped">
```
chrisma: Where is this `list` partial actually used? Is it needed? we already have `app/views/projects/_table.html.erb`

app/controllers/projects_controller.rb
BjoernDaase: Alright, removed the JSON part

app/views/projects/_list.html.erb
BjoernDaase: I removed it, it somehow appeared during a merge.

db/schema.rb
BjoernDaase: I removed the changes.

app/views/projects/_table.html.erb
```diff
@@ -1,25 +1,33 @@
...
+      <td>
+        <ul class="list-inline mb-0">
+          <% project.responsible_users.each do |user| %>
+            <li class="list-inline-item"><%= link_to user.name, user %></li>
+          <% end %>
+        </ul>
+      </td>
+      <td>
+        <% if project.responsible_users.include? current_user %>
+          <%= link_to fa_icon('trash'),
```
chrisma: The edit link is on the Project#show page. I'll move the delete button there too.

chrisma:
Ended up refactoring Project#show somewhat ¯\_(ツ)_/¯
Check it out

BjoernDaase:
@chrisma Thanks, the refactoring of the show page is great!
But is the 'Delete-Button' gone now? Or do I miss something?

chrisma:
> But is the 'Delete-Button' gone now? Or do I miss something?

The delete button is on the Project#show detail page, as deletion of projects is not a common need that needs to be available from the index. See https://github.com/hpi-swt2/vm-portal/pull/502/files#diff-2476f44a5f0802d146b181eeb56e8a12, line 33

BjoernDaase:
> > But is the 'Delete-Button' gone now? Or do I miss something?
> 
> The delete button is on the `Project#show` detail page, as deletion of projects is not a common task that needs to be available from the index. See https://github.com/hpi-swt2/vm-portal/pull/502/files#diff-2476f44a5f0802d146b181eeb56e8a12, line 33
> 
> image

Perfect!

PR #506, opened 2019-03-24
491/Conform-to-rails-standard

### LINKED ISSUE #491: Refactor the users and sudo_users to conform to rails standards
[//]: # (Please use this template for arbitrary tasks which aren't bugs or user stories. This issue type can be used for internal and development f)

**Description**
Currently, users and sudo_users don't behave like attributes of models, as they are saved immediately, not after validation and must always be assigned separately, instead of simply being assigned during update.

[//]: # (What this issue is about)

**Related Issues**
#401, #492 

[//]: # (Feel free to add new sections if required. Don't forget to add a priority as a label.)

MrModder:
Heavily refactor vms_controller

Closes #491


app/controllers/vms_controller.rb
```diff
@@ -179,29 +156,30 @@ def filter_vm_categories(user)
...
   end
 
   def not_enough_resources(exception)
     redirect_back(fallback_location: root_path)
     flash[:alert] = exception.message
   end
+
+  def assign_vm
+    @vm = VSphere::VirtualMachine.find_by_name(params[:id])
+    render(template: 'errors/not_found', status: :not_found) if @vm.nil?
```
chrisma: Should we also `redirect_back fallback_location: dashboard_path ...` here, instead of rendering the (pretty empty) error page?

spec/views/vms/edit.html.erb_spec.rb
```diff
@@ -3,28 +3,24 @@
...
-  end
   let(:configuration) do
-    config = double
-    allow(config).to receive(:description).and_return('hello world')
-    allow(config).to receive(:ip).and_return('127.0.0.1')
-    allow(config).to receive(:dns).and_return('8.8.8.8')
-    config
+    FactoryBot.create(:virtual_machine_config,
+                      name: 'VM',
+                      description: 'hello world',
```
chrisma: Is there a specific reason why the default params of the factory weren't used, i.e. `FactoryBot.create(:virtual_machine_config)`?

app/controllers/vms_controller.rb
chrisma: As this code is also used in other places, maybe we should discuss this in general

spec/views/vms/edit.html.erb_spec.rb
MrModder: Don't know, I just refactored this code from it's old state
But good point, I will set this to default

app/controllers/vms_controller.rb
MrModder: I had the same thought but then decided to use this methode because it was already in use elsewhere.

Using a redirect_back may however be advantageous on actions like update (POST, PATCH, etc.), in which a rendered page is not expected at all.
In these cases a redirect is a much better solution in my opinion

I would however want to use the vms_path as a fallback location instead of the dashboard_path as I think it is more strongly related with the vms controller.

PR #517, opened 2019-03-25
Faster loading of dashboard and vms/index page

BraunTom:


Gemfile
```diff
@@ -122,6 +122,12 @@ group :development do
   gem 'binding_of_caller'
   # Continous execution of unit tests
   gem 'guard' # https://github.com/guard/guard
+  # Benchmarking
+  # Supporting gem for Rails Panel (Google Chrome extension for Rails development)
+  gem 'meta_request' # https://github.com/dejan/rails_panel
+  # code profiler for MRI Ruby
+  gem 'ruby-prof' # https://github.com/ruby-prof/ruby-prof
```
chrisma: Should `ruby-prof-rails` be used instead? https://github.com/tleish/ruby-prof-rails

app/api/v_sphere/folder.rb
```diff
@@ -83,7 +90,11 @@ def create_vm(cpu, ram, capacity, name, cluster)
 
       vm_config = creation_config(cpu, ram, capacity, add_prefix(name), cluster.networks.first)
       vm = @folder.CreateVM_Task(config: vm_config, pool: cluster.resource_pool).wait_for_completion
-      VSphere::VirtualMachine.new vm
+      VSphere::VirtualMachine.new vm, folder: self
+    end
+
+    def equal?(other)
```
chrisma: `equal?` has a specific purpose in Ruby, i.e. to

> determine object identity (that is, a.equal?(b) if and only if a is the same object as b)

(http://ruby-doc.org/core-2.6.2/Object.html#method-i-eql-3F)

Therefore "Unlike ==, the equal? method should never be overridden by subclasses"

Most likely, `eql?` would be a better fit here

Gemfile
BraunTom: I tried to install it but got an error and Ruby-prof is good enough

app/api/v_sphere/folder.rb
BraunTom: changed it to eql?

Gemfile
```diff
@@ -122,6 +122,12 @@ group :development do
   gem 'binding_of_caller'
   # Continous execution of unit tests
   gem 'guard' # https://github.com/guard/guard
+  # Benchmarking
+  # Supporting gem for Rails Panel (Google Chrome extension for Rails development)
+  gem 'meta_request' # https://github.com/dejan/rails_panel
```
BjoernDaase: Do we really want to merge the `meta-request` and `ruby-prof` gems into `dev` or should we remove them before merging?

app/api/v_sphere/virtual_machine.rb
```diff
@@ -339,7 +352,7 @@ def request
       @request ||= Request.accepted.find_by name: name
     end
 
-    private
+    #private
```
BjoernDaase: Why is the `private` commented? Where do we need them to be non-private?

db/schema.rb
```diff
@@ -30,9 +30,9 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "vsphere_root_folder", default: ""
-    t.string "puppet_init_path", default: "/"
-    t.string "puppet_classes_path", default: "/Name"
-    t.string "puppet_nodes_path", default: "/Node"
+    t.string "puppet_init_path"
```
BjoernDaase: Due to @chrisma we should not have db changes in MRs.

Gemfile
BraunTom: The loading time is better but not optimal so when we need to benchmark later we have the gems already installed

app/api/v_sphere/folder.rb
```diff
@@ -11,17 +11,24 @@ class Folder
...
+      @parent = parent
+      @folder_name = name
     end
 
-    def parent
-      @folder.parent
+    def parent(lookup: true)
+      if @parent.nil? && lookup
+        parent = @folder.parent
+        parent.nil? ? nil : VSphere::Folder.new(parent)
```
MrModder: We should cache the parent here, e.g.
``` ruby
@parent = parent.nil? nil : VSphere::Folder.new parent
```

app/api/v_sphere/virtual_machine.rb
```diff
@@ -88,16 +88,21 @@ def self.user_vms(user)
...
-      if /^\d{8}_vm-/.match? @vm.name
-        @vm.name[12..-1]
+      return @name unless @name.nil?
+
+      @vsphere_name ||= @vm.name
+      if /^\d{8}_vm-/.match? @vsphere_name
+        @vsphere_name[12..-1]
       else
-        @vm.name
+        @vsphere_name
```
MrModder: Again, update the cache here (@name)

MrModder: And two lines above

app/api/v_sphere/virtual_machine.rb
```diff
@@ -147,11 +156,11 @@ def reboot_guest_os
     # We do not provide a power_state? method which just returns a boolean, because vSphere can internally handle
     # more than just two power states and we might later need to respond to more states than just two
     def powered_on?
-      @vm.summary.runtime.powerState == 'poweredOn'
+      summary.runtime.powerState == 'poweredOn'
```
MrModder: 👍

app/helpers/vms_helper.rb
```diff
@@ -31,6 +31,15 @@ def status_for(vm)
     end
   end
 
+  def status_color(vm)
```
MrModder: 👍

PR #518, opened 2019-03-26
Fix(Footer): Less HTML

chrisma:


chrisma:
Classic example of a badly written test case here: `spec/features/requests/accept_reject_request_spec.rb:49` fails, as it no longer finds the string `Info`, which was only ever contained in the footer's "More Information" heading...

BjoernDaase:
> Classic example of a brittle test case here: `spec/features/requests/accept_reject_request_spec.rb:49` fails, as it no longer finds the string `Info`, which was only ever contained in the footer's "More Information" heading...

So we should just remove this test and add another one which checks that the links are available?

chrisma:
> So we should just remove this test and add another one which checks that the links are available?

Well, the test `accept_reject_request_spec.rb` only ever passed because of a string in the footer, that is unrelated to what is actually being tested. However, accepting and reject requests is the core workflow of this application and should be throughly tested. The test should be fixed so that it actually tests the assumed behavior.

BjoernDaase:
> > So we should just remove this test and add another one which checks that the links are available?
> 
> Well, the test `accept_reject_request_spec.rb` only ever passed because of a string in the footer, that is unrelated to what is actually being tested. However, accepting and rejecting requests is the core workflow of this application and should be thoroughly tested. The test should be fixed so that it actually tests the assumed behavior.

Oh, now I see what you mean :+1:

BjoernDaase:
@chrisma Should be fixed now.

spec/features/requests/accept_reject_request_spec.rb
```diff
@@ -36,17 +36,17 @@
...
         end
 
         it 'saves the rejection information' do
           request.reload
-          expect(request.rejection_information).to eq('Info')
+          expect(request.rejection_information).to eq('A rejection information')
         end
 
-        it 'shows rejection information' do
-          expect(page).to have_text('Info')
```
chrisma: We should (re)add a test that the rejection information is actually displayed to the user

BjoernDaase:
@chrisma I added a test for that, which is currently failing. The problem is, that clicking the "Reject" button currently also results in an acceptance of the request (on hart-dev).  
Unluckily, I can not debug this problem on my own machine, because I always get the following error when either creating, accepting or rejecting a request: 
photo5296515629358098994
@BraunTom Do you have an idea what might have happend here?

BjoernDaase:
> @chrisma I added a test for that, which is currently failing. The problem is, that clicking the "Reject" button currently also results in an acceptance of the request (on hart-dev).
> Unluckily, I can not debug this problem on my own machine, because I always get the following error when either creating, accepting or rejecting a request:
> photo5296515629358098994
> @BraunTom Do you have an idea what might have happend here?

Due to @BraunTom related to the form in form on the requests page (which is not valid https://stackoverflow.com/questions/555928/is-it-valid-to-have-a-html-form-inside-another-html-form).

chrisma:
One large form with two submit buttons, one with the value "accept", one with the value "reject".

Issue #510, opened 2019-03-24
Remove JSON REST interface


chrisma:
The project contains many traces of JSON API, which we do not use, e.g. `respond_to {format}` and jbuilder files,.

* Remove all JSON paths from controllers where they aren't needed (which should be the case for everything but `Notification#any?`)
* Remove unused `*.json.jbuilder` files
* Remove jbuilder dependency

chrisma:
See https://github.com/hpi-swt2/vm-portal/pull/520

PR #520, opened 2019-03-26
Remove json responses from controllers

chrisma:
Ref https://github.com/hpi-swt2/vm-portal/issues/510


app/controllers/operating_systems_controller.rb
```diff
@@ -19,39 +18,28 @@ def new
...
 
   # DELETE /operating_systems/1
-  # DELETE /operating_systems/1.json
   def destroy
     @operating_system.destroy
-    respond_to do |format|
-      format.html { redirect_to operating_systems_url, notice: 'Operating system was successfully destroyed.' }
-      format.json { head :no_content }
-    end
+    redirect_to operating_systems_url, notice: 'Operating system was successfully destroyed.'
```
BjoernDaase: Could we better use `deleted` instead of `destroyed` here (and in `app/controllers/request_templates_controller.rb` line 42)? `Destroyed` always sounds weird for me.

chrisma: Good point. Anyways, this should be standardized using a single string in I18n, which is used in all controllers.

Issue #519, opened 2019-03-26
Edit software of server


chrisma:
When a server object has software added to it (heading `Software Information`), the items cannot be edited on the Server#edit page.

![image](https://user-images.githubusercontent.com/1652117/55006168-99e81000-4fdd-11e9-9d31-7fef50feed49.png)

* Write a test that checks that software added to a server is displayed when the server is edited
* Allow the edit form to change the listed software

chrisma:
https://github.com/hpi-swt2/vm-portal/pull/521

PR #521, opened 2019-03-26
519/Edit Software of server

chrisma:
See: https://github.com/hpi-swt2/vm-portal/issues/519


BjoernDaase:
I think we should also give the possibility to either remove software when editing or at least delete the empty fields. Otherwise, they will grow over time without having content.

chrisma:
> I think we should also give the possibility to either remove software when editing or at least delete the empty fields. Otherwise, they will grow over time without having content.

Software can be removed by visiting the `Software#edit` page and removing the rendered string from the input. You are correct, empty form fields are currently not removed in the controller. I didn' feel like investing any more work into a "feature" that should have been a single text field.

chrisma:
I fixed it anyways

app/controllers/servers_controller.rb
```diff
@@ -23,114 +22,44 @@ def new
...
   # Use callbacks to share common setup or constraints between actions.
   def set_server
     @server = Server.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
   def server_params
-    params.fetch(:server, {})
+    # Remove empty software form fields
+    params[:server][:installed_software].reject!(&:blank?) if params[:server][:installed_software]
```
MrModder: Won't this crash if params[:server] is nil?
We should to check for this in the condition

chrisma: > Won't this crash if `params[:server]` is nil?

Yes. If no `server` param is given and this line wasn't there, there would be an error one line later in the `require` call: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require

chrisma: But good point, the original exception is neater. I've dug around in `ActionController::Parameters` and found an acceptable way of changing params after permit.

BjoernDaase:
@chrisma @MrModder I resolved the conflicts so this should be ready to merge now, right?

Issue #407, opened 2019-02-21
Show puppetscripts on VM detail page


BjoernDaase:
[//]: # (Lines starting with [//]: are considered to be comments. You do not have to delete those lines, they are not rendered upon release.)

**User Story**

As an admin I want to see the puppet script of a VM on its details page.

**Acceptance Criteria:**

- [ ] Show a puppet script of a VM on its detail page

[//]: # (Please don't forget to add the priority and complexity label to this issue.)

chrisma:
Currently, the puppet script is shown on the VM request detail page, however, it has no direct connection to VM requests

chrisma:
If possible, add some syntax highlighting

krichly:
Display it here:
<img width="1351" alt="bildschirmfoto 2019-03-07 um 11 10 04" src="https://user-images.githubusercontent.com/5772525/53949323-17e58500-40ca-11e9-9804-26c0517a3b7d.png">

krichly:
+ Remove from request

chrisma:
- [ ] Connect VMs to the requests that created them

- [ ] If there is a request, display that the VM was created through the Portal, otherwise provide a link to the vSphere center page for the VM, so that it can be administrated (this might require setting the URL to the vSphere center in the settings)

- [ ] Show the relevant information of the request on the VMs#show page

PR #523, opened 2019-03-26
Show Puppet Scripts on VMs#show

chrisma:
Ref https://github.com/hpi-swt2/vm-portal/issues/407


chrisma:
Checked with @mat156 , position of button for showing scripts is okay for now.

PR #529, opened 2019-03-27
515/better error handling

### LINKED ISSUE #515: Better error handling
- [ ] Create a new Rails.logger that creates a notification for each admin for each logged error
- [ ] Add a class to error notifications to render them in a different color

MrModder:
Closes #515


MrModder:
For some reason, the logger now seems to send all logs to the logfile, instead of STDOUT
I don't know why this is happening, but it's definitely not intended

MrModder:
> For some reason, the logger now seems to send all logs to the logfile, instead of STDOUT
> I don't know why this is happening, but it's definitely not intended

This is now fixed by using a Formatter

app/views/notifications/_list.html.erb
```diff
@@ -23,11 +23,13 @@
             <%= I18n.l notification.created_at,format: :short%> (<%= time_ago_in_words notification.created_at %> ago)
           </span>
         </h5>
-        <%= simple_format(notification.message).html_safe %>
+        <div class="<%= 'text-danger' if notification.error? %>">
```
chrisma: red text on the light blue background might look a little strange
image
Maybe we can use a different background color to indicate the nature of the notification as an error? Or add a small FontAwesome icon?

config/environments/production.rb
```diff
@@ -81,8 +81,8 @@
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
-  # Use default logging formatter so that PID and timestamp are not suppressed.
-  config.log_formatter = ::Logger::Formatter.new
+  # Use custom formatter to support better error handling.
+  config.log_formatter = HartFormatter.new
```
chrisma: Using the new formatter is part of the application, so it's not specific to an environment, but should always be available in all environments. The formatter should be set in a general configuration, e.g. `config/application.rb`

MrModder: Ah, I was searching for a shared application file, but didn't find it. So `config/application.rb` is it.

app/views/notifications/_list.html.erb
MrModder: I don't want to add special css for a new background color (bootstrap does not support light backgrounds by default, the danger-background is too strong).
I will look into using an icon then

MrModder: image
Using an icon looks fine to me.

chrisma: :+1:

chrisma:
Thanks for the changes.
I could imagine that if an error occurs, it might occur multiple times in succession when using the application. This would result in the admin's notification list filling up with the same error multiple times.

It might be useful to prevent this behavior, e.g.:
* by adding a counter to the notification (`Occurred X times`) when the notification is still unread, 
* or to check whether the last notification is the same and to only update the creation date of this last notification, instead of creating a new one

spec/helpers/hart_formatter_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  it 'notifies admins when an error is logged' do
+    expect do
+      Rails.logger.error('My Error')
+    end.to change(Notification, :count).by(User.admin.size)
+  end
+
+  it 'does not notify admins when the log is not an error' do
+    expect do
+      Rails.logger.info('My Info')
+    end.not_to change(Notification, :count)
```
chrisma: A test for checking whether errors are displayed differently than usual notifications would be nice, e.g. in `spec/features/notifications/notifications_spec.rb`

MrModder:
> Thanks for the changes.
> I could imagine that if an error occurs, it might occur multiple times in succession when using the application. This would result in the admin's notification list filling up with the same error multiple times.
> It might be useful to prevent this behavior, by e.g.:
> 
> adding a counter to the notification (Occurred X times) when the notification is still unread,
> or checking whether the last created notification is the same as the one being created currently and only updating the creation date of this last notification, instead of creating a new one

I opted for the counter version. I think its more transparent.
This now also works for all notifications, not just error notifications.

image

app/helpers/hart_formatter.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Use default logging formatter in production mode so that PID and timestamp are not suppressed.
+class HartFormatter < Rails.env == 'production' ? ::Logger::Formatter : ActiveSupport::Logger::SimpleFormatter
```
chrisma: Should this be `ActiveSupport::Logger::Formatter`?

app/models/notification.rb
```diff
@@ -2,5 +2,26 @@
 
 class Notification < ApplicationRecord
   belongs_to :user
+  enum notification_type: { default: 0, error: 1 }
   default_scope { order(read: :asc, created_at: :desc) }
+
+  def ==(other)
```
chrisma: What is the benefit of using `==` here? The only time this function is used (that I can find) is in the `last_notification == notification` comparison (`app/models/user.rb:59`). This can be replaced with `last_notification.duplicate_of(notification)`, which is arguably clearer.

config/application.rb
```diff
@@ -16,6 +17,9 @@ class Application < Rails::Application
     # Load files in lib
     config.eager_load_paths << Rails.root.join('lib')
 
+    # set correct formatter for better error handling
```
chrisma: What is a `correct` formatter? What is `better error handling`? Explain why this formatter is needed.

spec/models/user_spec.rb
```diff
@@ -114,4 +114,45 @@
...
+
+    it 'notifies slack' do
+      allow(user).to receive(:notify_slack)
+      notify_user
+      expect(user).to have_received(:notify_slack)
+    end
+
+    it 'does not create multiple duplicate notifications' do
+      3.times { notify_user }
+      expect(user.notifications.size).to be == 1
```
chrisma: `.to eql 1` ?

MrModder: According to: [Equality matchers](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/equality-matchers)
You should use expect(a).to be == b if you want to check if a == b is true.
With numbers I would always use == and not eql?, so from my viewpoint this is the correct matcher

chrisma: `a.eql?(b)`: a and b have the same value
`a == b`: a and b have the same value _with type conversions_

Both are probably okay here. However, I would suggest:
`expect(a).to eq(b) # passes if a == b` which reads a little better, imho.

app/models/user.rb
```diff
@@ -51,16 +52,17 @@ def self.search(term, role)
   end
 
   # notifications
-  def notify(title, message, link = '')
-    notify_slack("*#{title}*\n#{message}\n#{link}")
+  def notify(title, message, link = '', type: :default)
+    last_notification = notifications.first
```
chrisma: not `notifications.last`?

MrModder: `notifications.first` is correct, Notifications are ordered by descending created_at order.

See models/notification.rb:
``` ruby
default_scope { order(read: :asc, created_at: :desc) }
```
So notifications with the newest ("largest") timestamp are first

app/models/user.rb
```diff
@@ -51,16 +52,18 @@ def self.search(term, role)
...
-    notification.user_id = id
-    notification.read = false
-    notification.link = link
-    notification.save # saving might fail, but there is no useful way to handle the error.
+    if notification.duplicate_of last_notification
+      last_notification.update(count: last_notification.count + 1)
+    else
+      notify_slack("*#{title}*\n#{message}\n#{link}")
+      NotificationMailer.with(user: self, title: '[HART] ' + title.to_s, message: (message + link).to_s).notify_email.deliver_now if email_notifications
+      notification.save # saving might fail, but there is no useful way to handle the error.
```
chrisma: > but there is no useful way to handle the error.
Well... we do have this `HartFormatter` that can log errors, which is introduced in this PR...

MrModder: That would not make sense, when saving the notification fails, we would log an error which would create a new Notification, which would most likely also fail to save, which would cause another error log...

We could write this in the comment however

app/models/user.rb
chrisma: Thanks for the info!

chrisma: https://github.com/hpi-swt2/vm-portal/pull/529/commits/2dd51a1a95e24e7b76abbf80c54c1d7aedec4e49

app/models/user.rb
chrisma: Oh, good point! I actually misread that line as somehow having to do with the `NotificationMailer` failing...
Thanks for adding the comment, that clears things up nicely.

spec/models/user_spec.rb
MrModder: I think `expect(a).to eq(b)` is too similar to `expect(a).to eql(b)`. The one letter difference can be easily overlooked. So I do not want to encourage usage of `.to eq` anywhere.
Especially as there usually is no a.eq?(b) method, so I would not know what `expect(a).to eq(b)` does. 

`expect(a).to be == b` may be a bit of an unusual syntax, but it clearly shows that == is used for the equality check. Which is the appropriate operation for equality checks on numbers.

chrisma: Good points. `eq` is the standard throughout the current code base.

spec/models/user_spec.rb
```diff
@@ -114,4 +114,45 @@
...
+
+    it 'does not notify slack for duplicate notifiations' do
+      allow(user).to receive(:notify_slack)
+      3.times { notify_user }
+      expect(user).to have_received(:notify_slack).once
+    end
+
+    it 'does create notifications if the duplicates are one minute apart' do
+      notify_user
+      new_time = Time.now + 61 # seconds
```
chrisma: This "magic number" (i.e. the 60 seconds in the model) might be better suited for the AppSettings.

chrisma: https://github.com/hpi-swt2/vm-portal/issues/543

PR #532, opened 2019-03-31
401/user management for servers

### LINKED ISSUE #401: user management for servers
**User Story**

Copy the user management functionality from the vms page and integrate it on the server pages.  

**Acceptance Criteria:**

- [ ] I can manage the users on servers

MrModder:
Closes #401 
Also fixes branching in git and reading of puppet scripts


app/models/server.rb
```diff
@@ -15,6 +15,10 @@ class Server < ApplicationRecord
   validates :ram_gb, numericality: { greater_than: 0 }
   validates :storage_gb, numericality: { greater_than: 0 }
 
+  MAX_NAME_LENGTH = 20
```
chrisma: Can we add a comment (e.g. a link to the vSphere docs) where this hard limit is documented?

MrModder: I'm actually not sure where this limit originates from, vSphere/Puppetscripts, etc.
As Servers are actually not vSphere objects (or are they?) they should not have this restriction anyway.
In any case, this code is mostly WIP, as its almost entirely copied from VMConfigs, I still have to remove all business logic that does not match servers.
Thank you for commenting anyway, I will look into it when I got the basic user management up and running

chrisma: > As Servers are actually not vSphere objects

Correct.

> I will look into it when I got the basic user management up and running

Thanks!

MrModder: > Can we add a comment (e.g. a link to the vSphere docs) where this hard limit is documented?

I actually could not find any reason why the name should be length-restricted.
Neither the vSphere-Docs, nor the puppet docs mention any such restrictions
I asked @BjoernDaase who previously was a PO, and he says it was a requirement by the customer, but he does not remember why.

I removed the length restriction and lower-case-only restriction from servers, as they are not a technical necessity. I left them in place for VMs though.

app/models/server.rb
```diff
@@ -27,4 +32,102 @@ class Server < ApplicationRecord
...
+    @users
+  end
+
+  def all_users
+    (users + sudo_users + responsible_users).uniq
+  end
+
+  # This makes a few things easier, for example when adding arrays together
+  def responsible_users
+    responsible.nil? ? [] : [responsible]
```
BraunTom: What is repsonsible? If it is an array there is an array in an array. Is that intended?

app/views/servers/show.html.erb
```diff
@@ -45,6 +45,27 @@
     <%= render 'templates/hardware_table', hardware: @hardware %>
   </div>
   <div class="column float-right half-screen-box">
+    <table class="table table-borderless" style="width: 100%">
+      <tbody class="table-active">
+      <tr>
+        <th colspan="2" ><h4 class="mb-0">Users</h4></th>
+      </tr>
+      <% (@server.all_users).uniq.each do |user| %>
```
BraunTom: all_users are already unique and I think you do not need the parens

app/models/server.rb
```diff
@@ -27,4 +32,104 @@ class Server < ApplicationRecord
...
+    end
+    @users -= @sudo_users
+    @sudo_users -= responsible_users
+  end
+
+  def convert_to_user(user_or_id)
+    user.is_a? User ? user : User.find(user_or_id)
+  end
+
+  def commit_message(git_writer)
```
chrisma: This exact code is also found in `app/models/virtual_machine_config.rb` and `app/models/request.rb`. Maybe it could be be moved to a shared helper?

app/models/server.rb
```diff
@@ -27,4 +32,104 @@ class Server < ApplicationRecord
...
+    else
+      ''
+    end
+  end
+
+  def node_script
+    Puppetscript.node_script(name, (sudo_users + responsible_users).uniq, all_users)
+  end
+
+  def write_puppetscripts(git_writer, name_script, node_script)
```
chrisma: Can this be deduplicated with `app/models/virtual_machine_config.rb` as well?

app/models/server.rb
```diff
@@ -27,4 +32,17 @@ class Server < ApplicationRecord
...
+  after_initialize :assign_defaults
+
+  # There is actually only ever one responsible user
+  # This method is a convenience accessor to allow for easier array concatination
+  # It is also recommended to use this method instead of :responsible because we might want to allow multiple responsible users in the future.
+  def responsible_users
+    responsible.nil? ? [] : [responsible]
+  end
+
+  def assign_defaults
```
chrisma: Can we assign the default using ActiveRecord? E.g. disallow `null` values in the db column, set '' as the default?

chrisma: Add a comment on why this is needed and can't be done via ActiveRecord!

Issue #408, opened 2019-02-21
Branch selection


BraunTom:
Add branch selection to git setting on the settings page

chrisma:
Again, sensible defaults for this, e.g. `master` should be in the `seeds.rb`

Issue #514, opened 2019-03-25
Refactor parsing puppetscript


chrisma:
* Investigate if there is a Ruby lib for parsing puppetscripts
* Parse using regexes with context information in capture groups
* Don't rely on the line number of scripts
* Get rid of `node_file_correct?`

chrisma:
We could make this more generic by relying solely on a regex to extract the user names, regardless of which line they are in, similar to `\$admins\s*=\s*\[["|'](\S+)["|']\]` , see: https://rubular.com/r/4AJ5PUkYflmzcK

PR #535, opened 2019-04-02
Fix/git and puppetscripts

MrModder:
Some issues with git and puppet scripts were introduced with #408 and #447 this PR fixes them


lib/puppetscript.rb
```diff
@@ -30,14 +30,24 @@ def self.generate_user_array(users)
...
+    result &&=  contents.lines[4].strip.eql?('realize(Accounts::Virtual[$admins], Accounts::Sudoroot[$admins])')
+    result &&=  contents.lines[5].strip.eql?('realize(Accounts::Virtual[$users])')
+    result &&=  contents.lines[6].strip.eql?('}')
     result
   end
 
+  def self.read_node_string(script)
+    values = { admins: [], users: [] }
+
+    admins = script.lines[1][/\[.*?\]/].delete('"[]').split(', ')
```
chrisma: We could make this more generic by relying solely on a regex to extract the user names, regardless of which line they are in, e.g. `\$admins\s*=\s*\[["|'](\S+)["|']\]` , see: https://rubular.com/r/4AJ5PUkYflmzcK

chrisma: I've added this comment to https://github.com/hpi-swt2/vm-portal/issues/514

PR #539, opened 2019-04-03
Massively speed up tests

chrisma:
Generating a new SSH key for every user employed in tests is expensive

Test run time: 2:30min -> 0:40min


MrModder:
I LOVE IT! 😍

